PR #2, opened 2022-05-06
Support Andor Solis .ASC files

mzieg:
support .asc files from Andor Solis

### DIFF
--- a/enlighten/AbsorbanceFeature.py
+++ b/enlighten/AbsorbanceFeature.py
@@ -3,12 +3,17 @@ import math
 
 log = logging.getLogger(__name__)
 
-##
-# Thinking this doesn't need VignetteROI, because Transmission already does it.
 class AbsorbanceFeature:
+    """
+    Computes absorbance using Beer's Law.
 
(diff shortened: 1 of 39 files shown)


enlighten/AbsorbanceFeature.py
```diff
@@ -17,10 +22,14 @@ def __init__(self,
         self.marquee        = marquee                
         self.transmission   = transmission
 
-    ## 
-    # @returns True if successfully updated ProcessedReading
-    # @todo update with numpy?
-    def process(self, processed_reading, settings):
+    def process(self, processed_reading, settings) -> bool:
```
rc14193, line 25: If you import ProcessedReading and SpectrometerSettings then you can also type annotate the parameters. I'm hoping to work towards all type annotations. Along with helping readability, I think it would be nice to have the option of using [MyPy](https://github.com/python/mypy) in a thorough manner. We could use now but since there aren't many type annotated parts I don't feel it would be very useful yet.

rc14193 (approved):

PR #4, opened 2022-05-10
Despike fix

rc14193:


mzieg:
This is definitely better than before, but I'm not sure it's actually useful for most Raman products in its current form.

mzieg:
Let's leave this in plugins until we have a working algorithm.

PR #8, opened 2022-05-12
RamanID testing

mzieg:
Added quotes to support directories with spaces in the names (like "Program Files").  Also removed "SimpleRamanID" plugin, as it really didn't differ.


mzieg:
Ready for this to be reviewed

rc14193 (approved):

PR #14, opened 2022-06-07
auto-detect utf-8-sig encoding

mzieg:
- auto-detect utf-8-sig encoding
- default to basename for label when loading CSV with no metadata

### DIFF
--- a/enlighten/MeasurementFactory.py
+++ b/enlighten/MeasurementFactory.py
@@ -226,32 +226,34 @@ class MeasurementFactory(object):
 
         # peek in the file and guess at the format
         try:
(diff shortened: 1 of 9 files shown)


enlighten/MeasurementFactory.py
```diff
@@ -226,32 +226,34 @@ def create_from_file(self, pathname, is_collapsed=False, generate_thumbnail=True
 
         # peek in the file and guess at the format
         try:
+            encoding = util.determine_encoding(pathname)
```
mzieg, line 229: this is the line that matters

rc14193 (approved):

PR #28, opened 2022-07-27
Mzieg connect feedback

mzieg:
Allow display of error messages arising from device connection.

### DIFF
--- a/enlighten/Controller.py
+++ b/enlighten/Controller.py
@@ -681,6 +681,10 @@ class Controller:
         ########################################################################
 
         if device.is_andor:
+            # Note that currently we're doing this BEFORE determining if it's a 
+            # hotplug; ideally, we should only need to connect to the Cloud
+            # on hotplug.  
+
             # attempt to backfill missing EEPROM settings from cloud
             # (allow overrides from local configuration file)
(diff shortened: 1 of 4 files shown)


enlighten/Controller.py
```diff
@@ -586,24 +586,34 @@ def check_ready_initialize(self):
...
             else:
                 # didn't get settings so check for timeout
                 if device.connect_start_time + datetime.timedelta(seconds=self.spec_timeout) < datetime.datetime.now():
                     log.error(f"{device_id} settings timed out, giving up on the spec")
-                    device.disconnect()
-                    self.multispec.set_gave_up(device_id)
-                    self.multispec.remove_in_process(device_id)
+                    disconnect_device = True
 
+            if disconnect_device:
```
rc14193, line 613: I'm fine with this. Trying to understand though, what is the purpose of moving it outside of the if else and using a flag?

rc14193 (approved):

enlighten/Controller.py
mzieg: To avoid duplicate code.  There's now two places in the above logic which might trigger a disconnect sequence (either a False SpectrometerResponse or a None+timeout).

PR #31, opened 2022-07-28
Usb event based

rc14193:


enlighten/Controller.py
```diff
@@ -471,7 +471,7 @@ def connect_new(self, other_device=None):
         if other_device is not None and other_device not in self.other_devices:
             self.other_devices.append(other_device)
         self.bus.device_ids.extend(self.other_devices)
-        if self.bus.is_empty():
+        if self.bus.is_empty() and len(self.multispec.spectrometers.keys()) == 0:
```
mzieg, line 474: self.multispec.count() preserves encapsulation (callers outside the class shouldn't need to know about the internal dict)

mzieg (approved):

PR #34, opened 2022-08-11
DeviceFinderUSB and related USB-reconnection logic

mzieg:
Hi Evan,

This is not a cohesive set of known-good changes. This is a set of tweaks I made over a span of several days while testing different types of spectrometers, with different goals. I'm not sure all the changes are necessarily good ideas, or required for success.

HOWEVER, I definitely have more success with a simple WP-785 unit, turning the spectrometer's power on and off repeatedly, and having ENLIGHTEN automatically and successfully reconnect each time, than with our current baseline on main.

Therefore, I think this changeset does more good than harm...but it honestly needs a second pair of eyes, and would benefit some some hard questions "Why did you do this?" if something doesn't make sense.

### DIFF
--- a/enlighten/Controller.py
+++ b/enlighten/Controller.py
@@ -34,6 +34,7 @@ from wasatch import applog
 from wasatch import utils as wasatch_utils
 
 from wasatch.WasatchDeviceWrapper     import WasatchDeviceWrapper
+from wasatch.DeviceFinderUSB          import DeviceFinderUSB
 from wasatch.SpectrometerResponse     import ErrorLevel
 from wasatch.StatusMessage            import StatusMessage
 from wasatch.WasatchBus               import WasatchBus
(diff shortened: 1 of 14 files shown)


enlighten/Controller.py
```diff
@@ -548,7 +555,8 @@ def connect_new(self, other_device=None):
         if isinstance(new_device_id.device_type, BLEDevice):
             pass
         elif str(new_device_id.address) != '111111' and str(new_device_id.bus) != '111111':
-            new_device_id.device_type = RealUSBDevice(new_device_id)
+            new_device_id.device_type = RealUSBDevice(new_device_id) # MZ: I don't like this
```
rc14193, line 558: Seeing the comments in Wasatch.PY, I will work on refactoring this so it is a proper implementation for distinguishing real vs sim devices

rc14193 (approved):

PR #35, opened 2022-08-16
Device ID Refactor

rc14193:
Restructures device creation so that BLE is created from the Device ID and DeviceID class no longer carries the entire BLE class.

### DIFF
--- a/enlighten/BLEManager.py
+++ b/enlighten/BLEManager.py
@@ -10,7 +10,7 @@ from threading import Thread
 from PySide2 import QtCore, QtWidgets, QtGui
 from PySide2.QtWidgets import QApplication, QDialog, QMainWindow, QPushButton, QVBoxLayout, QLabel
 
-from wasatch.BLEDevice import BLEDevice
+from wasatch.DeviceID import DeviceID
 
 log = logging.getLogger(__name__)
 
@@ -102,8 +102,9 @@ class BLEManager:
 
     def perform_connect(self, btn, device):
         log.debug(f"called to perform connect on btn {btn} with text {btn.text()}")
-        self.ble_device = BLEDevice(device, self.loop)
-        ok = self.controller_connect(self.ble_device)
+        #self.ble_device = BLEDevice(device, self.loop)
+        self.ble_device_id = DeviceID(label=f"BLE:{device.address}:{device.name}")
+        ok = self.controller_connect(self.ble_device_id)
         self.ble_present = True
         if ok:
             self.ble_button.setStyleSheet("background-color: blue")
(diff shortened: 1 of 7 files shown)


enlighten/Controller.py
```diff
@@ -541,24 +543,15 @@ def connect_new(self, other_device=None):
...
+        self.bus.device_ids.remove(new_device_id)
+        if new_device_id in self.other_devices:
+            self.other_devices.remove(new_device_id)
 
         # attempt to connect the device
         if new_device_id.is_andor():
             self.marquee.info("connecting to XL spectrometer (please wait)", persist=True)
         else:
-            self.marquee.info("connecting to 0x%04x:0x%04x" % (new_device_id.vid, new_device_id.pid), persist=True)
+            self.marquee.info(f"connecting to {new_device_id}", persist=True)
```
rc14193, line 554: I should add as it's easy to miss, the marquee will now say <DeviceID {BLE | USB | MOCK} {INFO}>. Not sure if you want maybe the DeviceID and <> stripped.

enlighten/BLEManager.py
```diff
@@ -102,19 +106,18 @@ def ble_discover_click(self, btn, device):
 
     def perform_connect(self, btn, device):
         log.debug(f"called to perform connect on btn {btn} with text {btn.text()}")
-        self.ble_device = BLEDevice(device, self.loop)
-        ok = self.controller_connect(self.ble_device)
+        #self.ble_device = BLEDevice(device, self.loop)
+        self.ble_device_id = DeviceID(label=f"BLE:{device.address}:{device.name}")
```
mzieg, line 110: I look forward to getting one of these units to test with :-)  (An actual BLE SiG, though I realize I could do with an RPi now...)

enlighten/Controller.py
mzieg: I'm okay with the verbose device description for now.  We can do a "friendliness / simplification" pass later, when all the cool new stuff is fully tested and working...it's reasonable to default to over-informative whilst in R&D.

scripts/Enlighten.py
```diff
@@ -119,7 +119,7 @@ def run(self):
         # The sim_spec code is used for debugging with the test framework during peculiar issues
         #####
         self.controller.control_exit_signal.exit.connect(self.closeEvent)
-        #sim_spec = MockUSBDevice("SiG_785","EEPROM-EM-9c65d19f4c.json")
+        #sim_spec = DeviceID(label="MOCK:SiG_785:EEPROM-EM-9c65d19f4c.json")
```
mzieg, line 122: Thank you for maintaining even commented-out code :-)

mzieg (approved):
Awesome.

PR #36, opened 2022-08-31
WIP: Ext trigger fix

mzieg:
Creating for review purposes (not yet submitted for merge)

### DIFF
--- a/enlighten/Multispec.py
+++ b/enlighten/Multispec.py
@@ -447,7 +447,7 @@ class Multispec(object):
         self.check_hide_others.setEnabled(multi)
         log.debug("update_widget end (%s)", self.device_id)
 
-    def readd(self, device_id):
+    def readd_spec_object(self, device_id):
         log.debug(f"readding spec {device_id}")
         try:
             self.spectrometers[device_id] = self.reset_spec_objs[device_id]
--- a/README_CHANGELOG.md
+++ b/README_CHANGELOG.md
@@ -2,6 +2,8 @@
 
 # 3.2 Open-Source
 
+- 2022-08-31 3.2.23
+    - fix for external trigger erroring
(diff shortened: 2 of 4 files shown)


enlighten/Multispec.py
```diff
@@ -428,6 +447,15 @@ def update_widget(self):
         self.check_hide_others.setEnabled(multi)
         log.debug("update_widget end (%s)", self.device_id)
 
+    def readd(self, device_id):
```
mzieg, line 450: Is this method name a typo?  If it's meant to be other than "read", recommend a longer version.

enlighten/TimeoutDialog.py
```diff
@@ -0,0 +1,59 @@
...
+# Presented solution by imo interestingly showing the many layers of inheritance in pyside
+# all QWidgets fire show events, which can have a slot
+# all QObjects can run a local QTimer that can have a slot
+# Combine the two and you can have a timed dialog, or other widget really
+
+from fileinput import close
+import PySide2
+from PySide2.QtWidgets import QMessageBox, QMessageBox
+
+class TimeoutDialog(QMessageBox):
```
mzieg, line 11: Very nice class.

enlighten/Multispec.py
rc14193: I'll change it to readd_spec_object. I shouldn't be using this anymore in the latest CEChanges. This was originally present because of what you described to me as David's desire for it to "quietly" reset. The idea here being that the high level spec object can persist while everything beneath it explodes and can be reconstructed with the reset device. The latest changes I kick off our spectrometer like you would see in the connection loop because it was overall easier and easier to work with. I think it would be valuable to keep this with a comment as it maybe something worth working on later so that we do get a quiet reset.

PR #39, opened 2022-09-19
got plug/unplug behavior working with test 785-SR-OEM

mzieg:


mzieg:
It occurs to me that we are imposing multiple expectations and requirements on ENLIGHTEN's disconnect / retry / reset / reconnection logic that aren't all in sync, and possibly contradictory.

When I'm testing at a desktop, plugging and unplugging units by hand, I want ENLIGHTEN to quickly notice that the unit is gone, and cleanly remove it from the graph / UI.  If I then replug the unit, I want it to re-enumerate quickly and successfully.  I repeat this test several times.

In contrast, in EFT testing, we want units to either "stay connected" or "automatically disconnect, then reconnect" in response to electrical faults.  This is a somewhat different workflow, and it's possible that the two use-cases are attempting to re-use some of the same settings and messages (e.g. SpectrometerResponse(False, poison_pill=True)) for different purposes.

rc14193 (approved):

PR #43, opened 2022-09-27
Plugin warning

rc14193:


enlighten/Controller.py
```diff
@@ -1886,7 +1899,8 @@ def process_reading(self, reading, spec=None, dark=None, ref=None, settings=None
 
         # This should be done before any processing that involves multiple
         # pixels, e.g. offset, boxcar, baseline correction, or Richardson-Lucy.
-        self.vignette_roi.process(pr, settings)
+        if self.roi_enabled:
+            self.vignette_roi.process(pr, settings)
```
mzieg, line 1903: I haven't tried to track down where the issue is, but this isn't sufficient.  The attached screenshot shows the CSV files generated if I save a spectrum (with ROI on), then click the button to turn the ROI off, then save a second spectrum, and then export both together.  The three files generated show that the "vignetted" metadata does toggle from TRUE to FALSE between the two spectra, but all three files are saved starting at pixel 17 (ROI Pixel Start).
image

mzieg (changes requested):

mzieg (changes requested):
Could you fix it so that a mixture of vignetted and unvignetted measurements will generate an export with all pixels (as required to show unvignetted spectra)?

I'm ambivalent on whether the vignetted columns should be _blank_ (,,) or _zero_ (,0,) for the "vignetted" pixels (or even -1, an impossible "raw" value).  I guess I don't much care, but would vote for blank.

mzieg (approved):

mzieg:
merging so I can test release candidate

PR #46, opened 2022-10-06
UI changes

rc14193:
UI overhaul for enlighten 4.0 release


enlighten/Authentication.py
```diff
@@ -121,6 +123,14 @@ def login(self):
 
         self.gui.colorize_button(self.button_login, self.level != self.BASIC)
         self.update_widgets()
+        if self.level > self.BASIC:
+            self.combo_views.addItem("Factory")
+        else:
+            try:
+                self.combo_views.removeItem(self.combo_views.findText("Factory"))
+            except:
+                log.error("Couldn't remove factory screen on deauth")
```
mzieg, line 132: Kept reading "de-auth" as "death" :-)

enlighten/BasicWindow.py
```diff
@@ -22,22 +25,29 @@ def __init__(self, title, headless):
...
     def create_signals(self):
         class ViewClose(QtCore.QObject):
             exit = QtCore.Signal(str)
         self.exit_signal = ViewClose()
 
+    def reconfigure_layout(self):
+        if self.size().width() < 1000:
+            self.ui.frame_scopeSetup_spectra.hide()
+        else:
+            self.ui.frame_scopeSetup_spectra.show()
```
mzieg, line 49: Elegant and simple.  Nice.

enlighten/BatchCollection.py
```diff
@@ -222,6 +223,11 @@ def __init__(self,
         self.spinbox_laser_warmup_ms       .valueChanged .connect(self.update_from_widgets)
         self.spinbox_collection_timeout    .valueChanged .connect(self.update_from_widgets)
 
+        # disable scroll stealing
+        for key, item in self.__dict__.items():
+            if key.startswith("cb_") or key.startswith("rb_") or key.startswith("spinbox_"):
+                item.installEventFilter(ScrollStealFilter(item))
```
mzieg, line 229: Dieter will thank you...

enlighten/BusinessObjects.py
```diff
@@ -423,6 +422,7 @@ def create_rest(self):
             parent                      = ctl.form,
 
             button_login                = sfu.pushButton_admin_login,
+            combo_views                 = sfu.comboBox_view,
```
mzieg, line 425: small nit: elsewhere we call it combo_view (singular).  Let's be consistent.

enlighten/Controller.py
```diff
@@ -128,6 +127,11 @@ def __init__(self,
         self.set_all_dfu            = set_all_dfu
         self.headless               = headless
         self.spec_timeout           = 30
+        self.form                   = form
+
+        if form is None:
+            log.error("Got a None value for form. Cannot start without QResources")
```
mzieg, line 133: > Cannot start without QResources

This describes many projects...

enlighten/Controller.py
```diff
@@ -1162,19 +1175,21 @@ def make_shortcut(kseq, callback):
...
-        make_shortcut("F2",     self.page_nav.set_technique_scope)
-        make_shortcut("F3",     self.page_nav.set_technique_raman)
-        make_shortcut("F4",     self.page_nav.set_technique_transmission)
-        make_shortcut("F5",     self.page_nav.set_technique_absorbance)
+        # views
+        make_shortcut("F1",     self.page_nav.set_view_hardware)
+        make_shortcut("F2",     self.page_nav.set_view_scope)
+        make_shortcut("F3",     self.page_nav.set_view_raman)
+        make_shortcut("F4",     self.page_nav.set_view_transmission)
+        make_shortcut("F5",     self.page_nav.set_view_absorbance)
```
mzieg, line 1183: We may revisit some of the non-Raman ones later, but good for now

enlighten/Controller.py
```diff
@@ -1162,19 +1175,21 @@ def make_shortcut(kseq, callback):
...
-        make_shortcut("Ctrl+1", self.page_nav.set_operation_mode_setup)
+        #make_shortcut("F6",     self.page_nav.set_operation_mode_setup)
+        make_shortcut("Ctrl+1", self.page_nav.set_operation_mode_raman)
+
+        #make_shortcut("F7",     self.page_nav.set_operation_mode_capture)
+        make_shortcut("Ctrl+2", self.page_nav.set_operation_mode_non_raman)
 
-        make_shortcut("F7",     self.page_nav.set_operation_mode_capture)
-        make_shortcut("Ctrl+2", self.page_nav.set_operation_mode_capture)
+        make_shortcut("Ctrl+3", self.page_nav.set_operation_mode_expert)
```
mzieg, line 1192: Not sure about the shortcuts.  I kind of like having Ctrl+1 be Setup View, and Ctrl+2 be Scope View (or vice-versa).  (Or just Ctrl+1 be "Toggle View between Scope and Setup").  Soft clay either way.

enlighten/Controller.py
```diff
@@ -1219,17 +1234,17 @@ def save_current_spectra(self):
         This is a GUI method (used as a callback) to generate one Measurement from
         the most-recent ProcessedReading of EACH connected spectrometer.
         
-        Originally there was no thought of Measurements "knowing" what technique
+        Originally there was no thought of Measurements "knowing" what view
         was in use when they were created.  However, we (currently) only want the
         ID button to show up on Raman measurements, so...let's see where this goes.
         """
-        technique = self.page_nav.get_current_technique()
+        view = self.page_nav.get_current_view()
```
mzieg, line 1241: I think this should maybe stay Technique; see comments in Measurement.py.

enlighten/Graph.py
```diff
@@ -199,7 +199,7 @@ def set_x_axis_label(self, text, locked=False):
         self.plot.setLabel(text=text+"<br>", axis="bottom")
         self.x_axis_locked = locked
 
-    ## when the Technique changes, update axis as appropriate 
+    ## when the View changes, update axis as appropriate
```
mzieg, line 202: When the Mode changes (Raman vs Non-Raman).

enlighten/Graph.py
```diff
@@ -276,7 +276,7 @@ def enable_wavenumbers(self, flag):
 
     ##
     # Only sets the "intention" to use the specified axis label; in reference-
-    # based techniques, don't actually switch to the target axis until processing
+    # based views, don't actually switch to the target axis until processing
```
mzieg, line 279: We need to figure out a consistent way to say this.  There might need to be a new "Nomenclature" section in the manual, and README.  In my mind:

- "Mode" is a way the _software_ behaves.  There will be a significant and growing set of on-screen options which vary between these three labeled sets: Raman, Non-Raman, and Expert.  
- Now, the fact that I named a "feature set" to "Raman" doesn't mean that "Raman" _isn't also a spectroscopic technique.  It is._  "Raman" will be offered as a convenience "application mode" (in which it shows all Raman-relevant widgets and features), but it also, scientifically and in our comments / documentation, remains a "Technique".  Basically though, at this time, we don't offer any techniques _other_ than Raman while in "Raman Mode", so "Raman" is being used both as a Mode (in page navigation) and as a technique (in internal comments).
- Raman is a non-reference technique.  It is a spectroscopic technique which does not use a reference.
- All reference-based techniques (Refl, Trans, Abs, Irrad etc) will be offered in the  "Non-Raman mode".  However, there will be some non-Raman techniques which don't use references.  So basically, we can't use either Mode or View as a surrogate to decide whether the current technique needs a reference.  
- We can assume that if we're in Raman mode, then we are in a non-reference technique (because Raman mode only supports Raman technique); however the converse is not true, because for instance "Emission" or "Fluorescence" are non-Raman techniques which don't use a reference.
- I don't remember what point I was trying to make, other than that we'll continue to tweak some of these comments.

enlighten/Measurement.py
```diff
@@ -714,7 +714,7 @@ def get_metadata(self, field):
         if field == "blank":                     return self.settings.eeprom.serial_number # for Multispec
         if field == "note":                      return self.save_options.note() if self.save_options is not None else ""
         if field == "temperature":               return self.processed_reading.reading.detector_temperature_degC if self.processed_reading.reading is not None else -99
-        if field == "technique":                 return self.technique
```
mzieg, line 717: I think probably all of these should have stayed "Technique."  In the UI, we're changing the "Technique" comboBox to a "View" comboBox.  However, we really do still need to keep track of Technique — it's still a valid concept, and it is what Measurement needs to know.

If we don't already have one, SpectrometerApplicationState should have a .technique member which can be an enum from the set { NONE, RAMAN, REFLECTANCE_TRANSMISSION, ABSORBANCE, COLOR, FLUORESCENCE, RELATIVE_IRRADIANCE } (for now).  All of those but "RAMAN" would be selectable from a new "Technique" widget in the ControlPanel.  The Technique widget should only be visible in Non-Raman and Expert Modes; when in Raman mode, the technique should automatically be set to Raman.  And though this slightly complicates things, if in Expert mode, then the comboBox should probably allow Raman to be selected as well.

Sorry for not making this clear as we continue to think-through what this design might involve.

enlighten/RamanIntensityCorrection.py
```diff
@@ -61,7 +61,7 @@ def set(flag, tt):
             return set(False, "Raman Intensity Correction requires an SRM calibration")
         elif spec.app_state.dark is None:
             return set(False, "Raman Intensity Correction requires a dark measurement")
-        elif not spec.app_state.technique_name in ["Raman", "Offset"]:
+        elif not spec.app_state.view_name in ["Raman", "Offset"]:
```
mzieg, line 64: Please remove "Offset".

mzieg (changes requested):
This is a really good start.  

The main delta is I mis-communicated, or honestly hadn't really thought-through, that instead of moving from "Mode" and "Technique" to "Mode" and "View" (still a 2-dimensional navigation), we're actually moving from 2D nav to 3D.

We're basically splitting the old "Technique" list into one list of "Views" and a second list of "Techniques".  So  now we have "Mode" (Raman, NonRaman, Expert), "View" (Scope, Setup, Device, Log) and "Technique" ([Raman], Refl/Trans, Abs, Color, Fluor, Irrad, etc).

Key takeaways:
- "Raman Mode" is the one mode that implies a technique
- "View" should never actually change processing / math.  Only a change in Technique should change math.
- "NonRaman" and "Expert" Modes need to surface the Technique comboBox so the user can tell us what math to use.

scripts/Enlighten.py
```diff
@@ -122,6 +144,7 @@ def run(self):
         #sim_spec = DeviceID(label="MOCK:SiG_785:EEPROM-EM-9c65d19f4c.json")
         #self.controller.connect_new(sim_spec)
         # pass off control to Qt to manage its objects until application shutdown
+        self.splash.finish(self.controller.form)
```
mzieg, line 147: I wonder if we still need a showMaximized() here?

enlighten/common.py
```diff
@@ -2,62 +2,59 @@
 import os
 
 from enum import IntEnum
+from pyexpat import ExpatError
```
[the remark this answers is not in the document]
rc14193: I do not remember writing this import. Maybe I was typing something else and it auto completed and I forgot about it. I don't see it being used anywhere, so I will remove it.

rc14193:
> 

Yes that's fine with me. I think that will be good. The one caveat I have is with plugins. I don't see a reasonable way to accommodate multi-graph or pandas output plugins, the same being true for the display all with KIA, on small screens like VGA. It seems to me like that would just be too busy of a screen. We could maybe do a stacked widget instead of side by side, but that means only one item visible at a time, which isn't ideal.

mzieg (approved):
Recommend we hold-off on merging to main, but design looks good.  Well done!

PR #51, opened 2022-10-10
Mzieg raman shift correction

mzieg:
Fixed two bugs:

- Original code checked if delta was 'less than' MAX_WAVENUMBER_SHIFT, which didn't cover the case of the delta being strongly negative; correct comparison uses abs(delta)
- Original code (in Wasatch.PY) had a bug where we might use a floating-point pixel coordinate as a list index lookup (the value was actually integral, but the datatype was float64)

### DIFF
--- a/README_CHANGELOG.md
+++ b/README_CHANGELOG.md
@@ -2,6 +2,8 @@
 
 # 3.2 Open-Source
 
+- 2022-10-10 3.2.34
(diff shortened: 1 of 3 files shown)


rc14193:
Is there a commit missing? I don't see how an integral type is ensured? Does there need to be an int() call along with the abs() call?

rc14193 (approved):

PR #52, opened 2022-10-12
Laser ctrl changes

rc14193:


enlighten/LaserControlFeature.py
```diff
@@ -193,6 +185,10 @@ def disconnect(self):
     def refresh_laser_button(self):
         spec = self.multispec.current_spectrometer()
         enabled = spec.settings.state.laser_enabled if spec else False
+        if enabled:
+            self.button_toggle.setText("Laser Off")
+        else:
+            self.button_toggle.setText("Laser On")
```
mzieg, line 191: Actually, I really would prefer "Turn Laser On" and "Turn Laser Off".  When people read "Laser On" they don't necessarily know if that's "current status" or "what the button will do."  Note that some users (even some staff) are colorblind, so the CSS doesn't automatically help.

enlighten/StatusIndicators.py
```diff
@@ -144,6 +144,14 @@ def update_visibility(self):
                         lamp = "disconnected"
                         lamp_tt = f"laser disarmed (cannot fire)"
 
+            # default behavior should be to do the normal processing
+            # Multispec case here checks all and will overwrite if a laser is on
+            all_specs = self.multispec.get_spectrometers()
+            specs_lasers_on = [s.settings.state.laser_enabled for s in all_specs]
+            if any(specs_lasers_on):
+                lamp = "warning"
+                lamp_tt = "laser is firing"
```
mzieg, line 153: Technically this is un-doing some of the laser interlock monitoring from the block above. The new code is based on ENLIGHTEN's SpectrometerState.laser_enabled, which actually indicates whether the laser has been ASKED to fire.  If the key-switch on the back of the spectrometer has been turned, or if the interlock continuity circuit has been broken, then the laser won't actually be firing, even though state.laser_enabled is True.

However, not many released customer systems yet support laser interlock monitoring, and I agree that this is an improvement over the current standard case.

mzieg (changes requested):

enlighten/StatusIndicators.py
```diff
@@ -123,7 +123,8 @@ def update_visibility(self):
             # Light Source (Lamp, Laser etc)
             ####################################################################
 
-            if settings.eeprom.has_laser:
+            all_specs = self.multispec.get_spectrometers()
+            if settings.eeprom.has_laser and len(all_specs) <= 1:
```
mzieg, line 127: Just an FYI...this is all inside a "spec is not None" so I don't think it's possible for len() to be <1 at this point.  Code's fine though.

mzieg (approved):

mzieg:
merging so I can test in the next release

PR #62, opened 2022-10-26
prefix and suffix in meta data and working export

rc14193:


enlighten/Measurement.py
```diff
@@ -329,6 +331,8 @@ def __init__(self,
         else:
             raise Exception("Measurement requires exactly one of (spec, source_pathname, measurement, dict)")
 
+        self.prefix = self.save_options.prefix() if self.save_options is not None else ""
+        self.suffix = self.save_options.suffix() if self.save_options is not None else ""
```
mzieg, line 335: As with the other PR, I'd probably move this under "if spec:" to avoid stomping over loaded files, and also default these to None in clear().

mzieg (changes requested):

enlighten/Measurement.py
rc14193: Is there a reason for None as opposed to "". I'm thinking "" so it will still be a string and shouldn't throw an error when writing. I think None may throw an error.

mzieg (approved):

PR #63, opened 2022-10-26
fix note overwrite on export

rc14193:


enlighten/Measurement.py
```diff
@@ -233,7 +233,8 @@ class Measurement(object):
                            'Battery %',
                            'Device ID',
                            'FW Version',
-                           'FPGA Version']
+                           'FPGA Version',
+                           'NOTE']
```
mzieg, line 237: To me this is weird, because "Note" is already in CSV_HEADER_FIELDS.  Was this a case-sensitivity issue?

enlighten/Measurement.py
```diff
@@ -714,7 +716,6 @@ def get_metadata(self, field):
         if field == "integration time":          return self.settings.state.integration_time_ms
         if field == "timestamp":                 return self.timestamp
         if field == "blank":                     return self.settings.eeprom.serial_number # for Multispec
-        if field == "note":                      return self.save_options.note() if self.save_options is not None else ""
```
mzieg, line 717: Ah.  Yeah, that was the issue :-/

enlighten/Measurement.py
```diff
@@ -233,7 +233,8 @@ class Measurement(object):
                            'Battery %',
                            'Device ID',
                            'FW Version',
-                           'FPGA Version']
+                           'FPGA Version',
+                           'NOTE']
     EXTRA_HEADER_FIELDS_SET = set(EXTRA_HEADER_FIELDS)
 
     def clear(self):
```
mzieg, line 240: I'd probably add self.none = None

mzieg (changes requested):

mzieg (approved):

PR #69, opened 2022-11-25
Virtual spec gui

rc14193:
As part of version one I want to merge in the option without virtual generation. I have some code to perform virtual spectra generation but have it disabled in Wasatch.PY as I wasn't satisfied with the results. I want to have something since non-synthetic is desired I thought best to merge that in while continuing to improve spectra generation.


enlighten/Controller.py
```diff
@@ -268,10 +269,9 @@ def __init__(self,
             }
 
         self.header("Controller ctor done")
-        self.other_devices = []
 
         # init is done so display the GUI and destory the splash screen
```
mzieg, line 273: destroy :-)

enlighten/MockManager.py
```diff
@@ -0,0 +1,157 @@
...
+from wasatch.MockUSBDevice import MockUSBDevice
+from wasatch.DeviceID import DeviceID
+
+log = logging.getLogger(__name__)
+
+class MockManager:
+
+    # mock_defaults will be available for generated spectra option
+    mock_defaults = {"WP-785-ER": ("WP_785_ER", "WP_785_ER.json"),
+                     "NIR1": ("WP_00998_NIR1", "WP_00998.json")}
```
mzieg, line 19: Eventually I'd like this to be dynamically generated from the set of installed files, but okay for now.

enlighten/MockManager.py
```diff
@@ -0,0 +1,157 @@
...
+        self.lamp_btn = lamp_btn
+        self.lamp_enabled = False
+
+        self.connect_btn.clicked.connect(self.connect)
+        self.lamp_btn.clicked.connect(self.toggle_lamp)
+        self.combo_compound.currentIndexChanged.connect(self.update_reading)
+
+        # Not satisfied with the current generated spectra
+        # For version one only doing files but not throwing
+        # away what I have, hiding for now
```
mzieg, line 56: good approach

enlighten/assets/example_data/mock_raman/mocks.json
```diff
@@ -0,0 +1,22 @@
+{
+  "Cyclohexane": {
+    "peak_location_cm": [ 392, 437, 804, 1031, 1165, 1273, 1358, 1453 ],
```
mzieg, line 3: These are close, but not the correct values?
Cyclohexane

enlighten/assets/example_data/mock_raman/mocks.json
```diff
@@ -0,0 +1,22 @@
+{
```
mzieg, line 1: I suggest that each record in this file get an optional "source_url" attribute that people can use to document where they got the peaks from.

mzieg (commented):
I look forward to testing this :-)

mzieg (approved):

samiebee43:
It is common to need some Spectra on the graph to test something on Enlighten.

- Clipboard
- File export
- Plugins

When I tested this, it was not clear to me how to get a fake spectra from the virtual device.

A static image of halogen lights would be sufficient in many cases.

image

enlighten/BusinessObjects.py
```diff
@@ -1,3 +1,4 @@
+from asyncio.windows_events import CONNECT_PIPE_INIT_DELAY
```
rc14193, line 1: Not sure why this was added. Maybe a weird linter autoimport. This should be removed though. It prevents the linux version from running since it is windows specific

rc14193:
> It is common to need some Spectra on the graph to test something on Enlighten.
> 
> * Clipboard
> * File export
> * Plugins
> 
> When I tested this, it was not clear to me how to get a fake spectra from the virtual device.
> 
> A static image of halogen lights would be sufficient in many cases.
> 
> image

I'm always open to answering any questions. Were none of the folders in testSpectrometers in ENLIGHTEN functioning for you? To setup your own, there is a bit more information in the Wasatch.PY [Virtual](https://github.com/WasatchPhotonics/Wasatch.PY/blob/master/README_VIRTUAL.md) readme. For a static image of a halogen lamp you could for example delete the spectra in the SiG_785 readings folder and throw in an exported halogen lamp reading. This would be as an example minimal implementation. For greater accuracy you could export the eeprom of the spec you got the reading from and then create it's own folder in order to make a virtual spec of it.

samiebee43:
Thank you for that info @rc14193. I'm going to close this PR for now in favor of merging in the opposite direction. main has diverged a bit too much since virtualSpecGUI

PR #70, opened 2022-11-28
add possible fix for international computers

rc14193:
Noticed in the description of the error that there was never any issues with creating EnlightenSpectra and that the difference here is that EnlightenSpectra is created by the python code not the InnoSetup. Changed the InnoSetup to now copy the example plugins to the install directory and then the Enlighten python code will copy the examples to the user docs if it detects no plugin folder present. This means the behavior should match now. The only potential issue is that any expanded, modified, or new examples won't be copied over on an update since the plugins folder will still exist.

### DIFF
--- a/README_CHANGELOG.md
+++ b/README_CHANGELOG.md
@@ -1,8 +1,11 @@
 # Version History
 
 # 3.2 Open-Source
-
+- 2022-11-29 3.2.40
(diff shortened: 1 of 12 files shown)


mzieg (approved):
My only comment is that it might make more sense to do this in PluginManager rather than Configuration, but either way it looks fine.

PR #83, opened 2022-12-14
added thumbnail tooltip

mzieg:
Request from the floor


mzieg:
Screenshot 2022-12-14 at 2 32 06 PM

PR #93, opened 2023-01-31
fixed logo in light mode

mzieg:
Also fixed installer bug (I think)

PR #95, opened 2023-02-02
fixed logo in light mode + bootstrap activate

samiebee43:
This branch contains [PR#93](https://github.com/WasatchPhotonics/ENLIGHTEN/pull/93) along with some minor adjustments to allow me (Samie Bee) to test it in Windows 11.

Before the change, bootstrap.bat would crash, terminating the condo environment, and making it impossible to open Enlighten from source. This change allows the user to run `bootstrap.bat activate` to skip steps that may interfere with activating the conda environment.

I was able to visually confirm that the logo looks correct in light mode.

image


mzieg (approved):
I notice some scripts lost their execute privs...this is common when adding/committing from the Windows side rather than host OS.  It's not a problem...this happens in all our repos (chmod toggling back and forth)...people who run scripts know how to fix it when needed :-)
image

Issue #106, opened 2023-02-20
Batch Collection measurement times drift over several hours


mzieg:
The current BatchCollection code computes the "start of the next measurement" as an offset from the "start of the previous measurement:"

https://github.com/WasatchPhotonics/ENLIGHTEN/blob/main/enlighten/BatchCollection.py#L556
https://github.com/WasatchPhotonics/ENLIGHTEN/blob/main/enlighten/BatchCollection.py#L591

It's good that it tries to achieve a full "start-to-start" period.  However, this approach has been found to drift over a period of hours.

It might be more accurate to instead base the "next start" time on current_batch_start_time + current_measurement_count x double(measurement_period_ms).

mzieg:
Further customer input suggests our current drift is on the order of 1 second / hour, which honestly isn't that bad.

samiebee43:
The amount of time it takes to schedule the next measurement after the dispatch of the previous one is roughly equal to the amount of time `save_complete` takes to execute. After an hour I would expect the drift to cause measurements to be late.

One possible solution is to have a a single start time and a variable that indicates how many measurements should have been made so far. At a frequency greater than the sample rate, perform the following check (pseudocode)

```
if int((time.now() - self.startTime) / deltaTime) > self.numberOfMeasurements:
    performMeasurement()
    self.numberOfMeasurements += 1
```

This way, the amount of time it takes for the code itself to execute is not excluded from our measurements.

samiebee43:
Instead of performing timing calculations, I'm approaching this by adjusting our use of [QTimer](https://doc.qt.io/qtforpython-5/PySide2/QtCore/QTimer.html)

- [ ] I've made some changes in b52425a which I expect eliminate some drifting. It does not handle absurd batch configurations as gracefully as main does. (For instance batch every 2000ms that consists of 3 x 1000ms measurements). Such configurations cause batches to fight over the variable current_measurement_count.

- [ ] The revision does not take the first measurement immediately like main does.

- [ ] Finally, timer_measurement should be changed to a recurring timer as well -- currently a sufficiently large batch would have drifting as well.

When these problems are resolved, the issue is ready to be closed.

samiebee43:
Absurd batch configurations, when the duration of a single batch is longer than the interval between batches, should be dealt with in one of the following ways.

![image](https://user-images.githubusercontent.com/124081765/221592380-e55c414e-dbda-4e10-853b-6de04671d644.png)

[...]

Issue #101, opened 2023-02-07
Add ✅ Laser Watchdog Enable


mzieg:
Currently, the laser watchdog (only supported with Series-XS spectrometers) supports a single spinbox, spinBox_ee_laser_watchdog_sec, with a range (0, 1200sec).

Please add an "enable" checkbox above the spinbox, and change the spinbox range to (3, 1200sec).

If the enable checkbox is OFF, LaserControlFeature should set the watchdog value to 0 sec and HIDE the spinbox.  (I'd rather not disable it, because it would still show a value, and the value would be non-zero, which would be confusing.)

samiebee43:
Completed in #102.

Notes. Changing a QTCheckbox state from within StateChanged signal creates undesirable behavior. This is a common use case when checkboxes have cancellable warning messages associated with them. The solution is to use the Clicked signal instead (which does cover "clicking" with the spacebar). I applied this change to both the watchdog and plugin checkboxes. It should be applied anywhere else where there is a checkbox and a warning message.

samiebee43:
The watchdog checkbox is now available in beta 4.0.5

Issue #99, opened 2023-02-06
Plugin execution msgbox shows dark in light mode


samiebee43:
![image](https://user-images.githubusercontent.com/124081765/216998546-b2ac440a-d1fb-4440-8055-0adf61bf7711.png)

AmanSal1:
in which file has this plugin warning window designed and coded ?

AmanSal1:
i tried looking for it but i didnt found it

AmanSal1:
How do we compile this code ? for test ?

AmanSal1:
I guess there is some issues with 2 files  :-

1.enlighten/assets/stylesheets/dark)/enlighten.css
2.enlighten/assets/stylesheets/light)/enlighten.css

In PluginController.py there is a function named "display_plugin_warning" which is used to generate the PluginWarning message box . This message box is not styled in any of the above two files thus causing an error that you have mentioned.

FIX:

QMessageBox {
    background-color: rgb(0,0,0);         # This should be included in the dark/ enlighten.css
  }
QMessageBox {
    background-color: rgb(255,255,255);    #This should be included in the light/enlighten.css
}

samiebee43:
Hi Aman. Thank you for your suggestions. I tried it out, but it's not yet working for me.
I will look more into it when I have a chance. Until then please be welcome to fork the repository and create a pull request 😁

AmanSal1:
can you plzz guide me on how to compile the code so i can test it locally on my machine ?

AmanSal1:
its bit confusing for me

samiebee43:
Same thing happens with some tooltips (clipboard)
![image](https://user-images.githubusercontent.com/124081765/220660996-cccff4ff-022d-43fe-b246-33ca41ad2759.png)

PR #115, opened 2023-03-17
Sbee patch

samiebee43:
**Overview.**

**Feb 8.** Includes a partial implementation for issue #99. It does not fix the issue, but it makes stylesheets available to the plugin controller who generates the warning message in question.

**Feb 16.** Separated Win10/Win11 build instructions. Added comment about checkbox bug addressed in [#101](https://github.com/WasatchPhotonics/ENLIGHTEN/issues/101#issuecomment-1422876556). Started Worek plugin.

**Feb 21.** Started improvements to README_PLUGINS.md: Answered the questions I had about plugins in an early section called "Quick Start". Added timeseries, dataframe, and fields to Worek. Includes partial implementation of #106.

**Feb 22.** Determined that callbacks are required for EnlightenPluginFields (undocumented).

**Feb 23.** Worek. Implemented callbacks for start/stop recording. Sample from scope for timeseries. Added slope start/end labels and vertical indicator lines. Indicator lines are later removed.

**Feb 27.** Change view -> technique. Reverse a token that was incorrectly changed via find-replace. This would cause an attribute error when loading any CSV file. 
Suppress exceptions from get_all_metadata which is called when generating a tooltip. These exceptions would prevent a CSV file from being added to the clipboard, even if it loaded successfully.
Started Ctrl-E to rename.

**Feb 28.** Completed Ctrl-E to rename.

**Mar 7.** Worek. Calculate values. Simplify interface. Use real wavelengths.

**Mar 9.** Worek. Configurable wavelengths, defaults on values from paper.

**Mar 15.** Worek. Remove Exception from getY. (Exceptions during plugin runtime cause them to go offline, requiring reconnect. So if the user sets an nonsense wavelength in the textbox for a second, we rather not crash)
.gitignore update
Patch. add try/except to our determine_encoding function, to add an 'iso8859_2' case for our friends across the pond.

**Mar 16.** Important. Using .get_processed() instead of .processed when dealing with ProcessedReading in Worek plugin. Otherwise the plugin will be seeing Emission values even when in Absorption mode.

**Mar 17th.** Documented `.get_processed()` and added quickstart to graphs in Enlighten plugins.

**Results.**

I would like to remove the partial implementations from February 8 and February 21. The first one is several changes that have no effect, and the second one removes a subtle drift from BatchCollect but introduces several behavioral changes (no immediate fire, undefined behavior for nonsense intervals). Instead we offer other solutions for precision batch collection.

Specifically I would like to exclude commits:
https://github.com/WasatchPhotonics/ENLIGHTEN/commit/a2ad43c40aa23807c7a05d14ec0224e0475387e3
https://github.com/WasatchPhotonics/ENLIGHTEN/commit/862e34acf55b34e8962d40cc2382a99f1646d422
https://github.com/WasatchPhotonics/ENLIGHTEN/commit/b1e32939fd6491dc11ce424f487812d2147f1ffa
https://github.com/WasatchPhotonics/ENLIGHTEN/commit/b52425a6d7246d652d1addeecef104e122c44a54

I will apply a new commit which reverses these changes called "undo stylesheet inject and BatchCollect".
Then I will make a commit that puts README_WIN10 and README_WIN11 up to date.


mzieg:
image

Hi Samie, when you have a PR that's ready for review, please add me as a reviewer :-)  I'll see if I can "default" myself in GitHub settings...

mzieg:
> I'll see if I can "default" myself in GitHub settings...

Hmm, I couldn't find a way to do that.

PR #120, opened 2023-04-04
4.0.7 installer issues

mzieg:
Looked at recent issues building installer.  Key findings:

- Not all our dependencies seem cool with the recently-released Python 3.11.  Locking Python down to 3.10.* helped.
- We had an old MacOS patch involving Matplotlib that suddenly starting causing issues on Windows.  The package isn't needed on Windows, so removed.
- Apparently PYTHONPATH gets defined in two places in bootstrap.bat; one of them was missing enlighten/assets/uic_qrc, which prevented enlighten_layout.py from finding grey_icons_rc.py

The updated code seems to properly build a working installer, posted at:

https://wasatchphotonics.com/binaries/apps/enlighten/beta/Enlighten-Setup64-4.0.7.exe

PR #121, opened 2023-04-05
Isolate tasks in bootstrap.bat

samiebee43:
Bring bootstrap debugging feature to mzeig-testing-4.0.6


mzieg (approved):
Changes makes sense.  My impression is that this will no longer have a "one-shot build installer" (equivalent to the old "installer" argument), but will rather require at least three invocations to match the old behavior (once with "rebuildall", then "pyinstaller", then "innosetup")?

scripts/bootstrap.bat
```diff
@@ -178,45 +233,48 @@ REM echo %date% %time% ======================================================
...
-if %errorlevel% neq 0 goto script_failure
-
-echo.
-echo %date% %time% ======================================================
-echo %date% %time% Logging Conda packages
-echo %date% %time% ======================================================
-echo.
-cmd /c "conda list --explicit" 
-if %errorlevel% neq 0 goto script_failure
+if "%install_python_deps%" == 1 (
```
mzieg, line 236: I think this needs to be "1"

scripts/bootstrap.bat
```diff
@@ -226,106 +284,99 @@ echo.
...
 echo PYTHONPATH = %PYTHONPATH%
 
-echo.
-echo %date% %time% ======================================================
-echo %date% %time% Regenerating Qt views
-echo %date% %time% ======================================================
-echo.
-sh scripts\rebuild_resources.sh
-if %errorlevel% neq 0 goto script_failure
+if "%regenerate_qt" == 1 (
```
mzieg, line 287: I think this needs to be "1"

mzieg (changes requested):

scripts/bootstrap.bat
samiebee43: it does

samiebee43:
> Changes makes sense. My impression is that this will no longer have a "one-shot build installer" (equivalent to the old "installer" argument), but will rather require at least three invocations to match the old behavior (once with "rebuildall", then "pyinstaller", then "innosetup")?

Yes. I didn't include a one-shot, since that never worked on my end, but it can be trivially added

scripts/bootstrap.bat
```diff
@@ -178,154 +248,142 @@ REM echo %date% %time% ======================================================
...
+)
 
-echo.
-echo %date% %time% ======================================================
-echo %date% %time% Regenerating Qt views
-echo %date% %time% ======================================================
-echo.
-sh scripts\rebuild_resources.sh
-if %errorlevel% neq 0 goto script_failure
+if "%regenerate_qt" == "1" (
```
mzieg, line 294: This needs to be %regenerate_qt% (closing marker)

mzieg (changes requested):

mzieg (approved):

samiebee43:
This PR and branch build-stage are no longer needed since #120 is up to date with sbee-build

PR #124, opened 2023-04-13
Added N2, O2

samiebee43:
Adding N2 and O2 to benigns.json is simple, so I wanted to make this change immediately available via a pull request, but there are a couple assumptions and areas that could undergo additional scrutiny. 

   1. - [x] I used common names Nitrogen, Oxygen to match other entries such as Isopropyl, Aspirin
2. - [x] I did ~not~ include "Alias" which could be "N2" and "O2" respectively
3. - [ ] I did not include "Type" to identify a peak as "Primary" or "Weak", modeled after Teflon, which is the only existing entry that has less than three peaks.
4. - [ ] There is not a sound entry. There are a few existing entries like this. Files like nitroglycerin.wav aren't mentioned in the local README.md, so I am not sure where to source more of them.

### DIFF
--- a/enlighten/assets/example_data/raman_id_signatures/benigns.json
+++ b/enlighten/assets/example_data/raman_id_signatures/benigns.json
@@ -3,6 +3,18 @@
     "Type": "ENLIGHTEN Raman Compound ID Database",
     "Compounds": {
 
+        "Nitrogen": {
+            "Peaks": {
+                "2328.4": {}
+            }
+        },
+
+        "Oxygen": {
+            "Peaks": {
+                "1552.9": {}
+            }
+        },
+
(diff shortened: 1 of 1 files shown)


samiebee43:
Feel free to check off the items that are good as they are and add comments for ones you want to expand on. I'll merge this once everything is resolved and checked.

Most likely, I imagine point number is two is the only thing we'll want to change right away. 

I want to point out that O2 is not perfectly synonymous with "Oxygen" which can bond with itself in [a number of ways](https://en.wikipedia.org/wiki/Allotropes_of_oxygen). So actually I'll default to including aliases O2 and N2 (added in last commit).

mzieg (approved):

PR #129, opened 2023-04-18
implement missing function causing KIA crash

samiebee43:
untested: I don't have KIA installed


samiebee43:
Since this is such a small change that couldn't possibly break anything (I hope I don't regret saying that) I think we should patch as a commit to main.
image

It's just this added function.. and this PR is not very useful anymore bc PageNav has had edits since

mzieg:
> Since this is such a small change that couldn't possibly break anything (I hope I don't regret saying that) I think we should patch as a commit to main.

One-line patches to main are the spice that make life interesting and occasionally exciting :-)

PR #137, opened 2023-05-05
show laser excitation spinbox in Expert mode

mzieg:
This used to be hidden through Authentication to prevent casual users from tripping over it, but now that we have a readily accessible Expert mode, this feature should be made easier to access.

### DIFF
--- a/enlighten/BusinessObjects.py
+++ b/enlighten/BusinessObjects.py
@@ -446,12 +446,10 @@ class BusinessObjects:
             parent                      = ctl.form,
 
             button_login                = sfu.pushButton_admin_login,
-            combo_view                 = sfu.comboBox_view,
+            combo_view                  = sfu.comboBox_view,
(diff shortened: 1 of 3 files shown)


enlighten/BusinessObjects.py
samiebee43, line 1: Need a spec with laser to test this

enlighten/BusinessObjects.py
```diff
@@ -503,6 +501,7 @@ def create_rest(self):
             button_toggle               = sfu.pushButton_laser_toggle,
             frame                       = sfu.frame_lightSourceControl,
             lb_watchdog                 = sfu.label_laser_watchdog_sec,
+            lb_excitation               = sfu.label_lightSourceWidget_excitation_nm,
```
samiebee43, line 504: Does this not have an event handler?

PR #138, opened 2023-05-08
fixed save on non-Raman

mzieg:
I received anecdotal reports from three different people having trouble saving spectra. I reproduced it on my laptop with a UV-Raman system and this resolved it.

### DIFF
--- a/enlighten/Controller.py
+++ b/enlighten/Controller.py
@@ -218,6 +218,7 @@ class Controller:
 
         # default to Scope view
         self.page_nav.set_main_page(common.Pages.SCOPE)
+        self.page_nav.set_view_scope()
 
         # configure acquisition loop
         self.setup_main_event_loops() # MZ: move to end?
--- a/enlighten/Controller.py
+++ b/enlighten/Controller.py
@@ -216,10 +216,6 @@ class Controller:
         self.business_objects.create_rest()
         self.graph.rehide_curves()
 
-        # default to Scope view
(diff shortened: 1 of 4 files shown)


enlighten/Controller.py
```diff
@@ -218,6 +218,7 @@ def __init__(self,
 
         # default to Scope view
         self.page_nav.set_main_page(common.Pages.SCOPE)
+        self.page_nav.set_view_scope()
```
samiebee43, line 221: ### Background info

`self.page_nav.set_main_page`
sets PageNav's stack_main index
which corresponds to QtObj:comboBox_view
image
This is the dropdown with options Scope, Settings, Hardware

`self.page_nav.set_view_scope`
sets PageNav's combo_view index
self.combo_view.setCurrentIndex
which corresponds to QtObj:stackedWidget_low
image
This is the widget that contains each of those screens listed in the dropdown.

These should not be independently settable outside of PageNavigation.

The solution is to put `self.page_nav.set_main_page` calls within PageNav's `set_view...` methods.

And actually that is already done.

### Conclusion

Line 220 is now redundant and should be removed

mzieg: Sounds good.  Please test on a non-Raman spectrometer to confirm.

mzieg: Actually, I just tested this and...it's not redundant.  It doesn't actually work without that line.

PR #140, opened 2023-05-09
change input params, and baseline estimation

samiebee43:
image

- baseline is defined as line passing through nearest point on spectrum to left and right

image
- the interpolated_baseline is the value that is subtracted, it matches the intersection of red and gold

image
- does not yet interpolate spectrum for intersection points

- area estimation is currently omitted (trapz(spectrum) - trapz(baseline))
- no decimals in pandas, not yet showing thousands separator
- green lines (inner-radius) removed

### DIFF
--- a/pluginExamples/Analysis/LocalBaseline.py
+++ b/pluginExamples/Analysis/LocalBaseline.py
@@ -116,3 +116,8 @@ class LocalBaseline(EnlightenPluginBase):
         self.metadata["OriginalPeak"] = peak
         self.metadata["PeakBaselineSubtracted"] = peak - interpolated_baseline
         self.metadata["Area"] = "--"
+
+        # keep input parameters in metadata
+        self.metadata["Left"] = self.get_widget_from_name("Left").value()
+        self.metadata["Right"] = self.get_widget_from_name("Right").value()
+        self.metadata["x"] = self.get_widget_from_name("x").value()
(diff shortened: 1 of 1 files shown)


pluginExamples/Analysis/LocalBaseline.py
```diff
@@ -20,13 +20,13 @@ def get_configuration(self):
...
+            initial = 10, minimum = 0, maximum = 10000, step = .2,
             datatype = "float", direction = "input"
-        )
-
+        )        
+        
         self.field(
-            name = "Baseline Extent",
+            name = "Right",
             initial = 10, minimum = 0, maximum = 10000, step = .2,
```
mzieg, line 30: Maybe set precision=1, since otherwise there's a frustratingly unreachable ".x0" at the end?

mzieg (approved):

pluginExamples/Analysis/LocalBaseline.py
samiebee43: EnlightenPluginFields with type="float" will always display two digits after the decimal no matter the value of step. The user can set the precision of 0.01 using the keyboard, also no matter the value of step. The step parameter only dictates how much the value changes using arrow-keys and scrollwheel.

Also, when adjusting the step param we want to be mindful that that will affect the scrollwheel speed. If we set it low then it will take a long time to move to the target value-- if we set it high then it will be easy to overshoot.

In any event, to remove the trailing 0, we would need to modify the bit of code that reads the EPF (dataclass) and generates widgets. QtSpinBox will have an option for number of digits after the decimal. I believe that is within `configure_gui_for_module` in PluginController.py

samiebee43:
image
illustrations persist when enabled and un-enabled

PR #153, opened 2023-05-16
conda deactivate + bootstrap ui

samiebee43:
### conda deactivate before activate

It gives an annoying deprecation message, but it avoids this scenario:

```
scripts\bootstrap activate
REM a lot of time passes
REM ...
REM did I activate? better do it again to be safe
run
```

And then it crashes because it was over activated.

Now this required activation step can be repeated to one's heart's content.

### bootstrap ui

After pulling updates I will try the lightest-weight way to run
```
scripts\bootstrap activate
run
```

If it crashes because of a UI element, then I'll figure I need to run `rebuild_resources.sh`
It's a lot easier to up arrow a couple times in the terminal and change 'activate' to 'ui', instead of typing out

```
sh scripts\rebuild_resources.sh
```

So... this is a convenience too.

### DIFF
--- a/scripts/bootstrap.bat
+++ b/scripts/bootstrap.bat
@@ -61,6 +61,12 @@ if "%1" == "test" (
 )
 
 REM DEFINE CUSTOM ACTION HERE
+if "%1" == "ui" (
+    set "regenerate_qt=1"
+    goto args_parsed
+)
+
+REM DEFINE CUSTOM ACTION HERE
 if "%1" == "custom" (
     set "rebuild_env=0"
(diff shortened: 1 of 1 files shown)


scripts/bootstrap.bat
```diff
@@ -60,6 +60,12 @@ if "%1" == "test" (
     goto args_parsed
 )
 
+REM DEFINE CUSTOM ACTION HERE
+if "%1" == "ui" (
+    set "regenerate_qt=1"
```
mzieg, line 65: I like that :-)

mzieg (approved):

Issue #160, opened 2023-05-23
Scrub defaults on new enlighten.ini


mzieg:
A 4.0.11 user seemed be defaulting to no file formats selected — CSV should probably default on. Also, two users voted for "rename files" to be defaulted on.

Also, export behaves weird if no file formats are selected.  I can see a use-case for what the save button "seems to do" when no formats are selected — it just adds the thumbnail to the Save Bar for visualization, which makes sense.  However, if you THEN check "CSV", and THEN do an export, I feel like it should still export the "format-less" measurements.  It doesn't, though -- it exports their metadata, but not the spectra.

PR #166, opened 2023-06-05
Fixes to "Save"

mzieg:
Changes:

- Configuration defaults work again (apparently broke at some point, including no longer defaulting SaveOptions.format_csv to True on new enlighten.ini)
- Warn user if "saving" spectra but no formats selected

### DIFF
--- a/enlighten/file_io/Configuration.py
+++ b/enlighten/file_io/Configuration.py
@@ -86,7 +86,7 @@ class Configuration(object):
 
         self.load_defaults()
         self.stub_missing()
-        self.stub_test()
+        self.stub_test() # MZ: I don't think this belongs in Configuration
 
         try:
             self.reload()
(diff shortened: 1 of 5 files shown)


enlighten/file_io/Configuration.py
```diff
@@ -341,11 +341,15 @@ def has_section(self, section):
             return section in self.defaults
 
     def has_option(self, section, option):
+        # any option which is defaulted will always indicate "has_option -> true";
+        # .ini file may override value, but can't change that there IS a value
+        if section in self.defaults and option in self.defaults[section]:
+            return True
```
mzieg, line 347: This was preventing "CSV" from being checked by default on new installations (no enlighten.ini file)

enlighten/measurement/Measurement.py
```diff
@@ -684,20 +684,30 @@ def rename_files(self):
...
             self.save_json_file()
+            saved = True
 
         if self.save_options.save_spc():
             self.save_spc_file()
+            saved = True
+
+        if not saved:
+            if self.measurements:
+                self.measurements.marquee.error("No save formats selected -- spectrum not saved to disk")
```
mzieg, line 710: This was leading users to believe they were saving data when they weren't

samiebee43: if, elif, else

mzieg:
this fixed #160

Issue #162, opened 2023-05-25
Export filename should use Prefix and Suffix


mzieg:
I actually thought I implemented this at one point, but maybe not.

Right now we have Prefix and Suffix fields on the GUI which simply prepend/append the strings directly to saved filenames.  I feel like if those fields have values, they should be applied to Export files as well; right now I think they aren't.

mzieg:
resolved in #168

PR #168, opened 2023-06-07
Fix various reported Save and Export issues

mzieg:
The use-case is to launch ENLIGHTEN, use "Load" to load several previously-saved single-Measurement CSV files, then clicking "Export".

Digging into the code, it looks like at some point ColumnFileParser started storing metadata internally as a list (presumably for use when loading multi-Measurement export files).  This broke single-file load.  Also there seemed to be confusion on field capitalization, and an assumption that metadata would always include wavecal coeffs.

The changes were tested by successfully loading several spectra collected over a span of days and re-exporting them.

### DIFF
--- a/enlighten/parser/ColumnFileParser.py
+++ b/enlighten/parser/ColumnFileParser.py
@@ -45,7 +45,7 @@ class ColumnFileParser(object):
 
     def parse(self) -> Measurement:
         # read through the input file by line, loading data locally
-        self.csv_loader.load_data()
+        self.csv_loader.load_data(scalar_metadata=True)
 
         # put loaded data into where it goes in ENLIGHTEN datatypes
         self.post_process_metadata()
@@ -61,17 +61,16 @@ class ColumnFileParser(object):
             processed_reading = self.processed_reading,
             save_options      = self.save_options)
 
-        # log.info(f"\nAfter creating measurement the wave coeffs are {m.settings.eeprom.wavelength_coeffs}\n")
-
-        # note that label often won't be what the user expected, because they
-        # may have editted it in the Capture Bar AFTER the file was created,
-        # and currently we don't re-save the file when the label is changed.
-        # (nor do we rename the file with the label, as that would rule-out
-        # many characters and syntax in the label)
         if "Label" in self.metadata:
             m.label = self.metadata["Label"]
         else:
+            # this was probably for an external script
             m.label = os.path.splitext(os.path.basename(self.pathname))[0]
+
+        # attributes of Measurement itself
+        for k in ["Note", "Prefix", "Suffix"]:
+            if k in self.metadata:
+                setattr(m, k.lower(), self.metadata[k])
 
         # this parser only loads a single measurement, so the file is renamable
         m.add_renamable(self.pathname)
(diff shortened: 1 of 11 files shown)


mzieg:
Fixes #162

enlighten/parser/ColumnFileParser.py
```diff
@@ -83,10 +83,20 @@ def parse(self) -> Measurement:
...
         if field not in self.metadata or self.metadata[field] is None:
             return 0
 
-        s = self.metadata[field].strip()
+        v = self.metadata[field]
+        if type(v) is list:
+            if len(v) == 1:
+                s = v[0]
+            else:
+                s = str(v)
```
samiebee43, line 95: This value `s` will never be safe to convert to float.

enlighten/parser/ColumnFileParser.py
```diff
@@ -83,10 +83,20 @@ def parse(self) -> Measurement:
...
+        else:
+            s = str(v)
+
+        s = s.strip()
```
samiebee43, line 99: not needed

enlighten/parser/ColumnFileParser.py
```diff
@@ -83,10 +83,20 @@ def parse(self) -> Measurement:
     # ##########################################################################
 
     def get_safe_float(self, field):
```
samiebee43, line 85: this method should be in Common.py and take the dictionary as an argument

mzieg: I'm hoping we can leave it local, now that it's smaller and simpler (less worthy of re-use)

enlighten/parser/ColumnFileParser.py
mzieg: simplified

mzieg:
found some bugs, taking back to draft

mzieg:
Expanded use-case:

- Take 2 spectra A.csv and B.csv
- exit ENLIGHTEN
- DISCONNECT SPECTROMETER
- relaunch ENLIGHTEN
- Load A.csv and B.csv
- Export A and B to make AB.csv
- (visually compare A, B and AB in Excel)
- exit ENLIGHTEN
- relaunch ENLIGHTEN
- load AB.csv
- re-export to create AB2.csv
- exit ENLIGHTEN
- (visually compare AB.csv and AB2.csv in Excel)

mzieg:
Hi Samie, just to forestall some obvious observations:

- ExportFileParser and ColumnFileParser clearly need to share more code, if not be merged into a single package
- Much of enlighten.file_io should probably follow CSVLoader into Wasatch.PY
- We should consider either changing the current CSV (individual and export) file formats, or adding new ones, that can be automatically loaded / generated through Pandas or similar, as long as they remain "intuitively useful" from Excel.  The current file formats were intended to be "easy for humans to use from Excel," but are a little clunky programmatically.  (Of course, that's what the JSON formats were provided for, so programmers wouldn't *need* to futz around with the CSVs.)

All of those are valid long-term goals.  At this point, I'm just trying to fix some clear but staightforward bugs in our current implementation.

enlighten/parser/ExportFileParser.py
```diff
@@ -414,6 +426,10 @@ def process_data(self, values):
                     if len(value) == 0:
                         continue
 
+                    # MZ: honestly not sure how to handle these
+                    if value == "NA":
+                        value = 0
```
samiebee43, line 431: [Suggestion] If it's a float you could use `float('NaN')`. Doesn't work for ints.

samiebee43 (approved):
I want to avoid using `setattr`, but I'll let it go.

[L199:ExportFileParser](https://github.com/WasatchPhotonics/ENLIGHTEN/blob/97519d5ee9e3bde6569cd7740d152927661300c6/enlighten/parser/ExportFileParser.py#L199)

It certainly makes the code prettier than the 8-line alternative, but it makes regex and static analysis harder.

samiebee43:
All non-resolved contentions, observations, whatever I put into issues which we can deal with later. Feel free to merge and close this thread.

Issue #164, opened 2023-06-02
Scope persistence


mzieg:
Scope value should be persisted between sessions.  I think right now it dynamically defaults to the midpoint of the active spectrometer, which is fine if a user has never defined a scope before, but if they *have* manually set one, I'd save that value through enlighten.ini.

Issue #158, opened 2023-05-19
Plugin Persistence


mzieg:
Obviously plugins are already free to create their own persistence frameworks, writing files wherever they like, in whatever format they like. However, it might be convenient for lightweight plugins to be able to benefit from the Configuration service ENLIGHTEN already has as a simple way to retain settings across application invocations (x-coordinates, etc).  We could use a sample plugin, like LocalBaseline, to demonstrate this.

Going a step further, sometimes we may want persisted settings to be "lockable", for instance so that an admin-user can setup the parameters for use with a plugin, and then those parameters will remain in use (visible but greyed-out) during user-level operations.  

A simple (admittedly crude) mechanism for this in my mind might be for the plugin to publish a textfield input, whose value is then compared against a digest in the plugin code.  If the "password" is correct, the inputs are unlocked and may be modified. (I would include the password in a comment in the Python code, as a teaching example to others in how the system works, and so ENLIGHTEN plugin developers could test it themselves.)

I recognize that "proper" authentication and persistence systems are a vast topic unto themselves, this just strikes me as a lightweight way to meet a current user request while leaving room for a more significant investment in time if and when need justifies.

mzieg:
This is now easily achieved through each plugin's self.ctl.config handle.

Issue #156, opened 2023-05-18
Multiple Cursors


mzieg:
The addition of "multiplexing" to the BaselinePlugin got me thinking about our whole Cursor feature.  Right now, we've got one Cursor, and you can turn it on or off.  It might be worthwhile just to replace the current [x] enable checkbox with [+] Add cursor, and let the user create as many as they like.  I might move the y-value display from the StatusBar up to the widget, and just show (x, y) tuples for each Cursor (along with a button to assign color).  Any plugin could then access the current list of Cursors and use them to provide on-screen "draggable" x-coordinate inputs to the plugin, which will honestly be a pretty common request.

samiebee43:
Love that I could distinguish this from a breaking bug due to our new tags :)

samiebee43:
Starting with some designs.

## Existing cursor x/y
x is highlighted orange, y is purple
![image](https://github.com/WasatchPhotonics/ENLIGHTEN/assets/124081765/00f4620e-4d3f-4a3a-8bf1-396037dce4ab)

## New location
![image](https://github.com/WasatchPhotonics/ENLIGHTEN/assets/124081765/468457ab-473f-44e1-b632-bd3a113e0d58)

## Mockup: Cursor default

![image](https://github.com/WasatchPhotonics/ENLIGHTEN/assets/124081765/295f519b-13de-44ff-9538-a0cd7e3d3931)
## Mockup: Cursor multiple

![image](https://github.com/WasatchPhotonics/ENLIGHTEN/assets/124081765/0aa70427-0351-4881-b281-d4fab2d07155)

## Decisions

We should decide if we still want checkboxes with add/delete. The advantage is that this way keeps track of the user's x-values like the current implementation does for the single cursor. The disadvantage is that it starts to clutter.

![image](https://github.com/WasatchPhotonics/ENLIGHTEN/assets/124081765/5be5ed1a-9dcf-4366-a2de-7fe3ef5ccf3e)

We may also consider using the 'eye-ball'. It's supposed to be a checkbox specific to visibility:

![image](https://github.com/WasatchPhotonics/ENLIGHTEN/assets/124081765/77007cc9-6158-4034-8cf8-5f61ea8d3166)

I'm showing two styles in the same screenshot, borrowed from gimp and blender.

## Enlighten eyeball

After writing the last section, I just noticed we currently use the eyeball in Enlighten, but not exactly correctly. 

[...]

PR #187, opened 2023-06-19
Sbee persist gain

### LINKED ISSUE #173: Persist gain_db in enlighten.ini

samiebee43:
Close #173

### DIFF
--- a/enlighten/scope/GainDBFeature.py
+++ b/enlighten/scope/GainDBFeature.py
@@ -70,7 +70,9 @@ class GainDBFeature(object):
         self.bt_dn.clicked.connect(self.dn_callback)
 
         # load gain_db from .ini
-        self.set_db(self.ctl.config.get_int("detector", "gain_db", 8))
+        ini_gain_db = self.ctl.config.get_int("detector", "gain_db", 8)
+        log.debug("Get gain from INI: %s", ini_gain_db)
+        self.set_db(ini_gain_db)
 
         self.update_visibility()
 
@@ -158,7 +160,8 @@ class GainDBFeature(object):
         self._quiet_set(self.spinbox, db)
         self._quiet_set(self.slider, db)
 
-        # TODO save gain_db to .ini
+        # save gain_db to .ini
+        self.ctl.config.set("detector", "gain_db", db)
 
     def up_callback(self):
         util.incr_spinbox(self.spinbox)
--- a/enlighten/scope/GainDBFeature.py
+++ b/enlighten/scope/GainDBFeature.py
@@ -124,9 +124,11 @@ class GainDBFeature(object):
         self.spinbox.blockSignals(False)
 
         # load gain_db from .ini, falling back to spec.settings.state (copied from EEPROM)
-        ini_gain_db = self.ctl.config.get_float("detector", "gain_db", now_db)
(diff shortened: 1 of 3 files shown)


enlighten/scope/GainDBFeature.py
```diff
@@ -121,31 +123,39 @@ def reset(self, hotplug=False):
...
+        self.multispec.set_state("gain_db", db)
+        self.multispec.change_device_setting("detector_gain", db)
+
+        # ensure both gain widgets are correct, without generating additional events
+        self._quiet_set(self.spinbox, db)
+        self._quiet_set(self.slider, db)
+
+        # save gain_db to .ini
+        log.debug("Save gain to INI: %s", db)
+        self.ctl.config.set("detector", "gain_db", float(db))
```
mzieg, line 155: The problem is, this would force every SiG to use the same gain.  Normally we store and load spectrometer settings on a per-serial number basis.

Settings are normally saved in Configuration.save_file, and loaded (this could arguably move to Spectrometer) in Controller.set_from_ini_file().

mzieg (changes requested):

enlighten/scope/GainDBFeature.py
samiebee43: > Normally we store and load spectrometer settings on a per-serial number basis.

Does this apply to #156, #158, #164 ?

mzieg: Good question.  I would say no, and here's my thinking:

- integration time and gain are very much "device settings" — they're set within the spectrometer, they have "startup" fields in the EEPROM to initialize their per-device settings, etc.  Any setting within ENLIGHTEN which more-or-less directly maps to a hardware setting or existing EEPROM field should probably be tracked per-device.

- plugins (#158) and the cursor (#156 and #164) feel, to me, more like "ENLIGHTEN settings." I think it's fair to treat those as application-level settings.  That said, this could be debated.

To me, the .ini format itself somewhat leans toward "simple settings" (integration time, gain, anything which needs to store exactly one scalar) being "convenient" for per-serial number store, while "complicated settings" where you have to store more than one value, and maybe include sets of values are simpler if given their own Configuration section.

Basically, I'm ambivalent on this.  I personally think it would be easier and more intuitive if Cursor and Plugins were application-wide, but if someone else had a strong opinion otherwise I could be persuaded.

samiebee43:
Need a second spec to test this.

Specifically, looking at multispec load from .ini -- does GainDBFeature.reset get called every time user changes current_spec?

mzieg (approved):

samiebee43:
## Acceptance Criteria

- (apply debug to show gain for FX2)
- Open Enlighten
- Set Gain to 32
- Close Enlighten
- Open it again
- Check that Gain is 32

This is a test after the change to be specific to serial numbers. I did not test this with multiple spectrometers connected.

enlighten/scope/GainDBFeature.py
mzieg: We should probably also persist Scan Averaging, and your question remains pertinent: should that be per-serial number or "application-wide?"  In this case I'm going to go with per-serial number, because on at least some spectrometers (SiG), that _will_ be an attribute written to the firmware via USB/BLE.  (And currently it's handled within WasatchDevice, again a per-unit association.)

PR #195, opened 2023-06-27
support processed_reading fields in note/prefix/suffix templates

mzieg:
We had been expanding {field} templates for note, prefix and suffix fields before the processed_reading was added to the Measurement, which is where many of the fields can come from.


enlighten/measurement/Measurement.py
```diff
@@ -318,6 +315,11 @@ def __init__(self,
             self.timestamp          = datetime.datetime.now()
             self.baseline_correction_algo = spec.app_state.baseline_correction_algo
```
mzieg, line 316: Technically we could have populated plugin_name here, as long as one was connected and enabled.

samiebee43 (approved):

PR #203, opened 2023-06-28
Sbee dev

### LINKED ISSUE #183: Invalid path shown next to Save .ini button
Invalid path shown next to Enlighten > Hardware > Save .ini File

image

Only after the save it updates:

image

The default message should be something along the lines of
`"calibration will be saved to %s" % self.pathname`
in Configuration.py

samiebee43:
## Changes

* Refactor GainDB
* Resolve #183 

## Acceptance Criteria

* Boot ✅✅
* Get Spectra ✅✅
* Switch to Hardware page, correct string shown next to [Save .ini File] ✅
* Debug Gain visibility always True, set gain persist after restart ✅✅
* Save measurement, to disk, and display in 'Clipboard/Measurements'  ✅
* Load measurement from disk ✅

Two checkmarks means another check after merging main


enlighten/BusinessObjects.py
```diff
@@ -663,13 +663,7 @@ def create_rest(self):
...
 
         self.header("instantiating GainDBFeature")
-        ctl.gain_db_feature = GainDBFeature(ctl = ctl,
-            bt_dn                       = sfu.pushButton_gain_dn,
-            bt_up                       = sfu.pushButton_gain_up,
-            label                       = sfu.label_gainWidget_title,
-            multispec                   = ctl.multispec,
-            slider                      = sfu.slider_gain,
-            spinbox                     = sfu.doubleSpinBox_gain)
+        ctl.gain_db_feature = GainDBFeature(ctl = ctl)
```
mzieg, line 666: I basically disagree with the premise that the application is improved by making everything global (every business object has access to basically everything in the application, including accessing every GUI widget through a flat namespace), but I'm willing to accept it.  You can quote all the PEPs you want — I just place a higher priority on encapsulation.  Anyway, I'm not fighting it, and I do believe that the primary maintainer of a codebase has the right to move things around to suit their preferences (I did).  I will try to learn and follow your style so we're contributing to a consistent design.

enlighten/scope/GainDBFeature.py
```diff
@@ -1,11 +1,11 @@
-from enlighten import util
+from .. import util
 import logging
+import enlighten
```
mzieg, line 3: What does this line add?

mzieg (approved):
I had to fix two of my own bugs to fix this (honestly not sure how that happened), but I agree it works well.

enlighten/scope/GainDBFeature.py
samiebee43: It enables static tracing using the annotation on ln43.

enlighten/BusinessObjects.py
samiebee43: I think of it as a lesser of two evils. I lean towards 'the ctl' way because it makes it easier to track references via grep or static analysis, and semantic changes can be performed faster.

enlighten/scope/GainDBFeature.py
mzieg: Ah...okay

samiebee43:
TODO: "persistent GainDB via .ini" >> changelog

Also this shadows the device's 'default GainDB' eeprom setting. So the only time it gets loaded from eeprom is the first time this spec is connected to a particular computer. This isn't necessarily a problem, as it saves us unnecessary writes to the EEPROM, but it can be categorized as kind of weird default behavior-- and it does not support gain persistence across multiple computers, obviously.

mzieg:
> changelog

It's in there.

> this shadows the device's 'default GainDB' eeprom setting

Yeah.  So, history: 

EEPROM availability: originally, Wasatch spectrometers lacked an EEPROM, so the .ini file was literally where such things were mastered. Then we added the EEPROM, but still had lots of spectrometers in the field without EEPROM, so we were kind of maintaining both. Then we dropped ENLIGHTEN support for the old (non-FID) spectrometers, so "nearly all" spectrometers now have EEPROMs (*XL).

EEPROM fields: the "startup_" EEPROM fields, supporting initial integration time, gain, and TEC setpoint are relatively new.  So we were still very much using the .ini to let users configure the "normal, intended" settings for that unit.  Now we have the startup fields, but as you say EEPROMs have a limited write lifetime.

Scalable configurability: we don't necessarily want to let all users configure all settings...some settings can lead to degraded measurement quality, others could potentially infringe safety, others could even damage equipment. Some users are highly-qualified scientist/engineers, some are station operators. The thought was that we would roughly group settings into 3 categories: 1. those anyone could configure in the .ini, 2. those OEM/advanced users could configure in the EEPROM ("edittable"), and 3. production/"everything".  And now we've stirred that pot again by adding "Expert Mode," which would be about a 1.5 (e.g. can fiddle excitation on-screen, but still not write EEPROM).

So yeah, this is a sliding scale across multiple dimensions with multiple vocal stakeholders, and we're trying to maintain a balance across different use-cases.

In this particular case, the EEPROM startup fields are arguably more important for people _not_ using ENLIGHTEN (OEM integrators who want the unit to startup and operate in a specific mode on power-up).

Issue #189, opened 2023-06-21
Analysis.StatsBuffer broken


mzieg:
StatsBuffer plugin no longer draws curves, probably due to these (note different objects):

  File "Z:\code\ENLIGHTEN\enlighten\Plugins\PluginController.py", line 1198, in handle_response
    log.debug(f"found undeclared curve {name}...adding. in_legend={series.get('in_legend', True)}")
AttributeError: 'NoneType' object has no attribute 'get'

  File "Z:\code\ENLIGHTEN\enlighten\Plugins\PluginController.py", line 1198, in handle_response
    log.debug(f"found undeclared curve {name}...adding. in_legend={series.get('in_legend', True)}")
AttributeError: 'numpy.ndarray' object has no attribute 'get'

Also displays this warning in .err:

pluginExamples\Analysis\StatsBuffer.py:152: FutureWarning: Unlike other reduction functions (e.g. `skew`, `kurtosis`), the default behavior of `mode` typically preserves the axis it acts along. In SciPy 1.11.0, this behavior will change: the default value of `keepdims` will become False, the `axis` over which the statistic is taken will be eliminated, and the value None will no longer be accepted. Set `keepdims` to True or False to avoid this warning.

mzieg:
Resolved in #208

PR #208, opened 2023-06-28
Stats plugin

mzieg:
- updated Stats to functional plugin
- improved default plugin plot x-axis

### DIFF
--- a/pluginExamples/EnlightenPlugin.py
+++ b/pluginExamples/EnlightenPlugin.py
@@ -90,7 +90,14 @@ class EnlightenPluginBase:
         """
         in_legend = True
 
-        if x is None: x = np.arange(len(y))
+        if x is None: 
+            unit = self.enlighten_info.get_x_axis_unit()
+            if unit == "nm" and len(y) == len(self.settings.wavelengths):
+                x = self.settings.wavelengths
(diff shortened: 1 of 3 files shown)


mzieg:
Resolves #189.

pluginExamples/EnlightenPlugin.py
```diff
@@ -88,9 +88,19 @@ def plot(self, y, x=None, title=None, color=None):
         @param title plot title, shown in legend
         @param color color of plot line
         """
+        if y is None:
```
samiebee43, line 91: Please move this check to the caller.
I'm planning to support `plot(title="something", color="orange")` as a way to add arbitrary entries to the legend.

It's useful in LocalBaseline where I want to have n lines, but only one legend entry per color:
```
self.plot(title='X', color='red')
for i in range(n):
    self.plot(x=[...], y=[..])
```

samiebee43 (changes requested):
See comment on line 91 of EnlightenPlugin.py

pluginExamples/EnlightenPlugin.py
mzieg: fixed

samiebee43 (approved):

Issue #161, opened 2023-05-25
Reimplement Thumbnail Tooltip Metadata


mzieg:
Right now, if you mouseOver a Thumbnail on the Clipboard, it gets a "huge" ToolTip providing all the Metadata for that Measurement. Some users like that, others vocally do not.

One option would be to add a button to the "Thumbnail button bar" under each Thumbnail, where clicking it would show/hide the metadata.  Or pop it out in a Dialog.

Or maybe we could "ctrl-MouseOver" or something?

Or add a configurable checkbox in Settings?

Or...whatever.  Different users like different things :-)  Just capturing for the backlog...

lenziwill:
We should consider ENLIGHTEN default settings, the settings that are active when the software is opened. For example, should ROI and SRM both be on when software is opened?

mzieg:
This was changed to WhatsThis.

PR #215, opened 2023-06-29
added min/avg/max to thumbnail tooltips

mzieg:
just seemed like a good idea


enlighten/ui/ThumbnailWidget.py
```diff
@@ -518,7 +518,11 @@ def disable_trash(self):
         self.button_trash.setVisible(False)
 
     def generate_tooltip(self):
-        tt = ""
+        # quick stats in first line
+        proc = self.measurement.processed_reading.processed
+        tt = f"Max {int(max(proc))}, Avg {int(sum(proc)/len(proc))}, Min {int(min(proc))}\n\n"
```
samiebee43, line 523: tt = ... if proc else ""

samiebee43 (changes requested):

enlighten/ui/ThumbnailWidget.py
mzieg: fixed

README_CHANGELOG.md
```diff
@@ -13,8 +13,10 @@
         - improved floating-point precision
     - renamed VignetteROIFeature to HorizROIFeature
     - persist gain_db by serial number
+    - added min/avg/max to thumbnail tooltips
     - plugins
         - updated Savitzky-Golay
+>>>>>>> main
```
samiebee43, line 19: merge conflict indicator

mzieg: fixed

mzieg:
I saw that #161 was asking if there was a compromise between always showing that huge mass of metadata when mousing-over as a ToolTip, and just disabling the feature (as some people think it's "too much.")  Turns out Qt actually has a [QWhatsThis](https://doc.qt.io/qt-6/qwhatsthis.html) feature specifically designed for this, so I took a whack at testing it.

For some reason my laptop isn't showing cursors when I take screenshots, and cursors are pretty integral to this, so I took a quick movie instead:

https://youtu.be/lnbQtACkg5E

For the record, I'm not completely sold on this implementation, but I think it's a step in the right direction.  We'd probably actually want to move the new WhatsThis button up next to the current Help button, and add more WhatsThis text across the GUI, and maybe play with the CSS and SVG a bit...but this works, and I think is arguably better than what we had before.

mzieg:
Resolved #161

Issue #198, opened 2023-06-27
Correct MeasurementFactory attribute view -> technique


samiebee43:
> That fixes the bug, but this attribute really shouldn't be called "view" at all. "Absorbance," "Transmission" etc aren't Views (screens), they're spectroscopic techniques. This used to be the case in 3.x (https://github.com/WasatchPhotonics/ENLIGHTEN/blob/3.2.42/enlighten/MeasurementFactory.py#L125), and it was incorrectly changed early in 4.x.

https://github.com/WasatchPhotonics/ENLIGHTEN/commit/526a30b88fcff3cf4616d63f43a8e3b30095a71e#r119708790

mzieg:
Addressed in #223

PR #224, opened 2023-07-04
Add filename template

### LINKED ISSUE #223: Filename templates
By user request:

- Add a Settings field for “filename template”, defaulting to “{measurement_id}-{serial_number}”
- Add new template macros for {date} (YYYY-MM-DD) and {integration_time_sec} (include 3 digits of floating-point precision)

mzieg:
Will resolve #223


mzieg:
Following discussion with customer, expanded the scope of this change a bit.  I'm kind of feeling my way, and don't want to let any one customer dictate our naming conventions, but trying to support multiple use-cases without drowning in complexity / permutations.

- Add new checkbox "use filename as label".  If checked, the on-screen label will always use the filename.  The label template field will be disabled, and batch collection will no longer add its suffix.
- renamed and moved the old MeasurementFactory.label_suffix to SaveOptions.multipart_suffix.  This is still only currently populated by BatchCollection, but we'll have other applications which collect a series of related measurements.
- Template changes
  - macros can include spaces (needed to access some metadata fields)
  - built-in {time} and {date} macros
  - built-in {integration_time_sec} macro (user request)
  - built-in {file_timestamp} giving filesystem-safe timestamp
- Usability fixes to ThumbnailWidget
  - Label field is wider, with smaller font
  - Can finish editing by re-clicking the pencil icon
  - Edits which don't change the text aren't acted on

Tested with this measurement label: {laser_power_mW}mW_exp{integration_time_sec}sec_{scans_to_average}scan_{note}_{file_timestamp}

Yielded this filename:
450.00mW_exp0.010sec_1scan_ball5mm_2023-07-05_11_23_31621440

I suspect the customer may still not like the periods...will iterate.

mzieg:
image
Wider label

image
New template field, new checkbox

samiebee43:
Resolve #198

samiebee43:
> I suspect the customer may still not like the periods...will iterate.

```python
locale.format_string("%.2f", 54.54)
```

mzieg:
> > I suspect the customer may still not like the periods...will iterate.
> 
> ```python
> locale.format_string("%.2f", 54.54)
> ```

I guess I didn't mean use of period vs comma for decimal separator (Japan does use period), as many people don't like any periods in filenames other than the final extension.  All I can say is that ```integration_time_ms``` will remain integral for people who don't like fractions...

enlighten/measurement/Measurement.py
```diff
@@ -445,18 +446,55 @@ def expand_template(self, template):
...
+            # note all date components are upper-case and times are lower-case 
+            # for consistency (may confuse C programmers, should make sense to 
+            # spectroscopists)
+            elif macro == "YYYY": value = self.timestamp.strftime("%Y")
+            elif macro == "MM": value = self.timestamp.strftime("%m")
+            elif macro == "DD": value = self.timestamp.strftime("%d")
+            elif macro == "hh": value = self.timestamp.strftime("%H")
+            elif macro == "mm": value = self.timestamp.strftime("%M")
+            elif macro == "ss": value = self.timestamp.strftime("%S")
+            elif macro == "ffffff": value = self.timestamp.strftime("%f")
```
samiebee43, line 488: This could be an anonymous dictionary expression:

```python
value = self.timestamp.strftime({
    "time": "%H_%M_%S",
    "date": "%Y-%m-%d",
    ...
    }.get(macro))
```

Pros:
* easier to read
* easier to add entries
* less prone to hard to spot logic errors: (elif->if, value=->value==)

Cons:
* more to difficult to understand
* last elif containing `fmt =` will need to be patch at end
* this format is not as flexible to arbitrary semantic changes

enlighten/measurement/MeasurementFactory.py
```diff
@@ -123,7 +122,7 @@ def create_thumbnail(self, measurement, is_collapsed=False):
             colors          = self.colors,
             stylesheets     = self.stylesheets,
             is_collapsed    = is_collapsed,
-            view            = measurement.technique,
```
samiebee43, line 126: I feel like I'm seeing some repeated diffs. Same with `if proc else ""`. 

Do you `merge main` or `merge bug-fix` often enough?

mzieg: Possibly not often enough.  Just did, though.

samiebee43 (approved):
Approved, but I'm weary that this feels like more unmanaged complexity.

I would be happier if this was more modular. `expand_template` should be in a place like Common and take a string and a dict as arguments. Especially if we're trying to rope BatchCollection into this.

That being said, I'm glad that `expand_template` is contained mostly in just the one function. This branch has several useful bug fixes and swiftly meets the needs of our customer. 

It's also a fairly big diff centered on a fairly critical region. So I don't want this to linger in PR space too long.

mzieg:
I don't disagree with any of that. I might be able to clean some of that up when I look at the open BatchCollection issue.

PR #225, opened 2023-07-05
Test SplashScreen messages

mzieg:
This doesn't do much useful on my dev laptop, but I wonder if it would be helpful on slower PCs?

Beyond low priority, was just goofing around while watching telly.

The main problem is that it's hard to predict whether light or dark text would be more generally useful.  Suppose we could use dark mode to tell, after we instantiated it...


mzieg:
image

samiebee43:
Line 84 of BusinessObjects, `header` add a `time.sleep`, to test this on a fast PC

samiebee43:
Ready for review?

mzieg:
Yeah sure.  Technically it doesn't add much value, but "it works" and provides "a place to put things" in the future, if and when we're interested in providing more user-visible feedback during startup.  (For instance, scanning each plugin, loading .ini persistence etc.)  And users like seeing "things happen" :-)

samiebee43 (approved):
Approve. Curious, why the new file splash.png ?

mzieg:
> Approve. Curious, why the new file splash.png ?

It added the grey bar underneath the logo, so the white text would have a consistent background.  I didn't want add the bar to the standard file because we use it for other things.

PR #227, opened 2023-07-05
show multiple local baselines. Not incl pd frame, center, meta

samiebee43:
This is an 'MVP' of LocalBaseline with multiplexing.

The full version is not inherently complicated, but the task is sitting on top of a problematic area within Enlighten Plugins (EPF, PFW, ...) Given timing constraints, I don't want a useful and simple to implement feature to be held back due to an ever pending architectural change. 

The difference is that this lighter version, dubbed sbee-min-baseline does not allow for runtime changes of the number of localbaseline's, whereas sbee-baseline would. Instead here it is set to exactly 4-- it can be changed in code.

## Progress

- [x] Show 4x (input x, Left, Right, Center) widgets
- [x] Center buttons work
- [x] Legend is free of clutter
- [x] Pandas frame shows 4x data
- [x] Metadata 4x

## Other changes

* Uncheck/Check 'Connected' now always reloads from file. This prevents needing to restart Enlighten so often

## Acceptance Criteria 

Used pyusb-virtspec for this. 

* Set x_0, x_1, x_2, x_3 to 200, 300, 400, 500. Confirm red lines are in correct locations. ✅
* Set Left/Right 0..3 to 50. Confirm that blue lines progressively space out. ✅
* Set x_i to 199, press center_i. Confirm it moves to top of peak. ✅
* Check Pandas frame output. ✅
* Check save file. ✅

## Screenshots

image
image

image
image


mzieg:
Testing this afternoon.

mzieg:
Down the road, I wonder if we might make the Pandas table more readable with auto-resizing columns:

https://stackoverflow.com/a/52665000

mzieg:
I was wondering if we could make the CSV data a little prettier, so was playing with this:
https://github.com/WasatchPhotonics/ENLIGHTEN/pull/235

While doing so, I ran into this:
Screenshot 2023-07-06 at 5 53 24 PM

After catching that, I caught this:
Screenshot 2023-07-06 at 5 56 58 PM

Right now it doesn't seem super-robust when changing the widgets while collecting live spectra.

mzieg (changes requested):
as above

mzieg:
> when changing the widgets while collecting live spectra

I wonder if this relates to using EnlightenPluginBase.get_widget_from_name() (which could be literally changing during processing of the request) vs EnlightenPluginRequest.fields, which I think takes a static scalar of each when the request is generated.

mzieg:
Short-form...you can't do math on "--"

samiebee43:
> Short-form...you can't do math on "--"

These last four commits make placeholder-ing more reliable.

> After catching that, I caught this: <screenshot>

This error is from mzieg-min-baseline. While it certainly might apply to this branch as well, from the surface it strikes me as not caused by this diff (maybe it is an existing edge case common to both). 

I would prefer it to be a post-merge issue barring a replication on sbee-min-baseline.

samiebee43:
OK looking into it, that error is just an obfuscation of `Arithmetic on "--"` which I handled in the last commit.

mzieg (approved):

mzieg:
Small GUI glitch which I don't consider a release blocker.  When typing a longer x value (like 1000cm⁻¹), it briefly responds and generates small baselines at 1cm⁻¹, then 10cm⁻¹, then 100cm⁻¹...and it doesn't always clear them. So I end up with little bits of orange on the screen that doesn't go away.  Fine for now.

mzieg:
image

PR #232, opened 2023-07-06
added virtspec to bootstrap

samiebee43:
## Local "Installation"
    
    pwd # path/to/Enlighten
    cd ..
    git clone git@github.com:WasatchPhotonics/pyusb-virtSpec.git

## Usage

Now "virtspec" can be added to the end of any bootstrap command.

```
scripts\bootstrap activate virtspec
```

To go back to using a real one:

```
scripts\bootstrap activate
```

## Note

Because of how it's kept outside of the Enlighten codebase, pyusb-virtspec will work with (probably) any version of Enlighten. To do so easily, you may want to patch the build script to your working tree:

```
git checkout <past version>
git checkout main scripts/bootstrap.bat
```

This process was helpful when I did a git bisect for the last bug fix.

### DIFF
--- a/scripts/bootstrap.bat
+++ b/scripts/bootstrap.bat
@@ -15,10 +15,16 @@ set "pyinstaller=0"
 set "innosetup=0"
 set "runtests=0"
 
+set "virtspec=0"
+
 REM Convoluted but safe way to generate an audible bell from a .bat
 REM without inserting control characters which confuse unix2dos etc.
 REM https://stackoverflow.com/a/64515648
 set "RING_BELL=echo x|choice /n 2>nul"
+
+if "%2" == "virtspec" (
+    set "virtspec=1"
+)
 
 if "%1" == "activate" (
     goto args_parsed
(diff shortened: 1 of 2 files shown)


mzieg (changes requested):
Could you copy-paste your nicely-formatted pull request description into a new README_VIRTSPEC.md?

scripts/bootstrap.bat
```diff
@@ -302,6 +311,19 @@ if "%install_python_deps%" == "1" (
...
+    echo.
+    echo %date% %time% ======================================================
+    echo %date% %time% * Appending PyUSB Virtual Spectrometer to PATH *
+    echo %date% %time% ======================================================
+    echo.
+    set PYTHONPATH=..\pyusb-virtSpec;%PYTHONPATH%
+    pip uninstall pyusb
+)
+if "%virtspec%" == "0" (
+    pip install pyusb==1.2.1
```
mzieg, line 324: If we're hardcoding pyusb version here, should that be in requirements.txt as well?

samiebee43:
> into a new README_VIRTSPEC.md?

OK for now, but I'm planning to move us off of README_X/Y/Z towards a more focused documentation hiearchy.

> nicely-formatted pull request description

Why thank you

scripts/bootstrap.bat
samiebee43: Yes absolutely. I would like to keep that change out of this PR though. 

Because while I am certain that v1.2.1 is exactly the right pyusb for everyone using Enlighten everywhere-- I won't necessarily have that much confidence for every module we are pinning, and if we have problems down the line it would be nice to revert the 'version pinning' commit without rolling back the 'virtspec' one.

PR #234, opened 2023-07-06
Default empty template fields

### LINKED ISSUE #233: template fields cannot be blank
If filename or label templates are blank, reset to default values (can't leave empty).

mzieg:
- resolves #233 
- starts a new pattern of encapsulating Configuration defaults within the owning class


samiebee43:
## Synopsis

First commit, looks good. Provides defaults and moves the default dictionary into SaveOptions to have easy access to `.DEFAULT_...`. I believe this is also the commit that resolves the aforementioned issue.

I disagree with [second commit](https://github.com/WasatchPhotonics/ENLIGHTEN/pull/234/commits/d2760198d4617ca4d6ce5a3d1ea94d8f6e3663f9). 

## Reasoning

I think the point was to let the Feature class have authority as to where it saves it's default configuration, but that's not a good thing. The naming `get_*` should be without side-effects. And we should avoid functions that overwrite data passed into them, especially at a fixed key like this.

```python
# it's obvious where the data came from and where it's saved
self.defaults["save"] = SaveOptions.get_default_configuration()
```

```python
# a getter with no return value ? which key did it get written to ?
SaveOptions.get_default_configuration(self.defaults)
```

## Extended rational / philosophy

Problems would ensue if this bit of code were scaled up:

```
d = {}
d[x_0] = y_0
d[x_1] = y_1
d[x_2] = y_2
A.get_default_configuration(d)
B.get_default_configuration(d)
C.get_default_configuration(d)
```

It's non-trivial to ensure that keys of A/B/C's implementation don't collide with `x_i`.

Alternatively, sticking to the previous way scales up easily.

```
d = {}
d[x_0] = y_0
d[x_1] = y_1
d[x_2] = y_2
d["save"] = A.get_default_configuration(d)
d["save"].update(B.get_default_configuration(d))
d["save"].update(C.get_default_configuration(d))
```

The point is not that we want to necessarily scale this piece up, rather to use how does this scale as a heuristic when deciding between two semantic equivalences. 

[aside]: Not about this PR, but this reasoning of 'premature scaling' could be (mis-)used to justify `setattr`'s and generative lambda stuff such as `process_func` -- to which I would say something like "except don't do that". Mainly because these methods seriously kill readability/static analysis/grep in exchange for pretty small bouts scalability.

mzieg:
> I disagree with second commit

Fixed.

samiebee43 (approved):

PR #235, opened 2023-07-06
testing LocalBaseline

mzieg:


samiebee43 (changes requested):
Some of this conflicts with my last commits handling "--". Please roll those back. These changes include everything except the definition and usage of `set_meta`.

The `set_meta` function is cool, but `i` should be a parameter, and the function should be called `set_meta_multiplex` or something like that. Also the function should be outside of any loops or other functions because it should have an explicit `self` parameter.

mzieg:
Closing as key changes already incorporated.

PR #236, opened 2023-07-07
Template format precision

mzieg:
I know we need to refactor this to manage complexity, but I realized this one small tweak would probably address a range of user requests.

- accepts templates of the form {gain_db:0.1f}
- everything after the colon is used directly as the format code
- only applied to "known float" values

### DIFF
--- a/README_CHANGELOG.md
+++ b/README_CHANGELOG.md
@@ -11,7 +11,7 @@
             - added filename template
             - macros now work in export filename components
             - support Reading attributes
-            - improved floating-point precision
+            - user-configurable floating-point precision
             - added date/time components ({YYYY}, {hh} etc)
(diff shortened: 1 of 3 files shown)


samiebee43 (approved):
Now with float precision this looks even more like a reimplementation of built in advanced format strings `"{var:...}".format(dict)`.

Still approved, for one key reason.

The python built in handles a single incorrect template less gracefully than your alg does:
```
"{{whoops} {correct1} {correct2}" --> ValueError
```

instead of
```
"{{whoops} {correct1} {correct2}" --> "{{whoops} value1 value2"
```

That being said, there's a limit to how many features from `.format` I'm willing to have separately reimplemented, and we're getting close to that point.

PR #251, opened 2023-07-27
Sbee dev

### LINKED ISSUE #250: Add FWHM to LocalBaseline

samiebee43:
* Close #250 
* Use both rows and columns in LocalBaseline

image

image

### DIFF
--- a/enlighten/scope/RegionControlFeature.py
+++ b/enlighten/scope/RegionControlFeature.py
@@ -49,6 +49,7 @@ class RegionControlFeature:
         log.debug(f"setting region to {n}")
 
         # update locally, to cause wavelengths / wavenumbers to be updated
+        # SB: needed? isn't this call generated using process_f in the next line ?
         spec.settings.set_single_region(n)
 
         # send downstream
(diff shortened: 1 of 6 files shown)


samiebee43:
Pretty sure this is ready for review. I think I left it as a draft because there's a couple of my snarky comments I was unsure about and I wanted to complete work on the RamanID plugin.

Since then I'm calling ready for review because (1) I don't want my FWHM implementation to go stale (2) We've since pivoted from the work that I thought would be completed in here (ask offline if you don't know what I mean) (3) The snarky comments are helpful in the long run.

enlighten/scope/RegionControlFeature.py
```diff
@@ -49,6 +49,7 @@ def region_callback(self):
         log.debug(f"setting region to {n}")
 
         # update locally, to cause wavelengths / wavenumbers to be updated
+        # SB: needed? isn't this call generated using process_f in the next line ?
```
mzieg, line 52: Possibly no longer needed.  This was probably written when ENLIGHTEN was multi-process, and there were literally different SpectrometerSettings objects in different processes.

mzieg (approved):

PR #254, opened 2023-08-12
New requested plugin features

mzieg:
Adding support for:
- autoloading a plugin at launch (cmd-line for now, could support .ini file later)
- ability for plugins to disable/hide standard ENLIGHTEN features

### DIFF
--- a/enlighten/Plugins/PluginController.py
+++ b/enlighten/Plugins/PluginController.py
@@ -244,7 +244,7 @@ class PluginController:
 
         # bindings
         self.button_process.clicked.connect(self.button_process_callback)
-        self.cb_connected.clicked.connect(self.connected_callback)
+        self.cb_connected.stateChanged.connect(self.connected_callback)
         self.cb_enabled.stateChanged.connect(self.enabled_callback)
         self.combo_module.currentIndexChanged.connect(self.combo_module_callback)
         self.combo_graph_pos.currentIndexChanged.connect(self.graph_pos_callback)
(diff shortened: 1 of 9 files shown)


enlighten/Plugins/PluginController.py
```diff
@@ -244,7 +244,7 @@ def __init__(self, ctl,
 
         # bindings
         self.button_process.clicked.connect(self.button_process_callback)
-        self.cb_connected.clicked.connect(self.connected_callback)
+        self.cb_connected.stateChanged.connect(self.connected_callback)
```
samiebee43, line 247: I think the checkbox stateChanged handler has a bug where some clicks won't invoke a message.

Something to the effect of: 
image

Details can be found in older commits on Enlighten related to watchdog and checkbox (if this is in fact the bug I'm thinking of)

mzieg: Good spot. It did work in my test across multiple clicks, whether through the GUI or the --plugin option.  However, it sounds like [others](https://stackoverflow.com/q/66697607) have had problems at least with older Qt versions.

I went ahead and changed it from stateChanged (which some say should only be used on tri-state checkboxes) to toggled, which seems the recommended signal for bools.  (I went away from clicked() because that definitely doesn't respond to setChecked()).  This still works in testing.

samiebee43 (changes requested):

mzieg:
Is that the change you requested?

samiebee43 (approved):

samiebee43:
Yes that works ty.

PR #258, opened 2023-08-17
removed stroker and dragster references

mzieg:


samiebee43:
Ah so the Wasatch.PY one was the right one. That might actually solve my Ubuntu issue

samiebee43:
Could you diff this with the Wasatch one?

If they are equal in contents and conceptually equivalent, lets consider removing this file from Enlighten.

samiebee43 (changes requested):
This looks good, but maybe we can go step further and remove the file completely

samiebee43 (approved):

PR #273, opened 2023-08-31
Updates to run from source, build .app and create .dmg installer on MacOS

mzieg:


mzieg:
Gah, this is frustrating.  It works fine from source, and it works ONCE from the compiled binary app...but only once.  On the SECOND call to DeviceFinderUSB.bus_polling, it dies — not on the call to usb.core.find(), but on the attempt to iterate the returned generator...

mzieg:
"Gah" fixed.

samiebee43 (approved):
Thank you for documenting

Issue #276, opened 2023-09-12
RamanIntensityCorrection usage confusing


mzieg:
To enable RamanCorrectionIntensity, you have to switch to Expert Mode.
However, per the following code in Controller.py, you have to be in Raman Mode for the feature to actually be applied:
```
1823             ####################################################################
1824             # Raman intensity correction
1825             ####################################################################
1826                          
1827             if self.page_nav.doing_raman():
1828                 self.raman_intensity_correction.process(pr, spec)
```
We need to think how to do this more elegantly.

mzieg:
Related, we should add a Raman "Technique" in Expert Mode—that would also have resolved this, and the current list of exclusively non-Raman techniques is misleading.

samiebee43:
Relabeling as [ 1 - broken ] because the most intuitive way to use it fails silently. Though this is kind of the edge of these two labels.

Also added it to milestone _Planned for 4.0.16_. The idea with the "Planned for..." milestones is that we will not release until all of such issues have been closed. In exceptional cases we may move issues to subsequent versions, but be mindful to notify downstream users in that case.

We may prefer to do Planned for's for semi-major versions in the future, so we can still release small patches.

samiebee43:
Is this closed by #277 ? @mzieg

mzieg:
Hi Samie, I believe #277 adequately resolves this for now.

PR #277, opened 2023-09-13
allow SRM in Expert Mode

mzieg:
- addresses #276


samiebee43 (commented):
Waiting for main to be up to date with 4.0.15 before merging anything related to 4.0.16.

That way we can visit a commit on main that has 4.0.15 code without any 4.0.16 code.

samiebee43 (changes requested):
Please add a comment explaining the nav

samiebee43:
I meant actually a three line comment on 1827 of Controller.

Thanks for taking the time to document PageNav class, but generally I want to avoid long discussions in comments, as they can distract from the code itself.

I prefer 5-10 lines for class strings, 3-7 for functions, and 1-3 for inline.

I've mentioned this elsewhere, I believe longer comments are more relevant in systems or driver code, such as Wasatch.PY -- I would like to keep longer descriptions or discussions about design in peripheral documents such as Google Drive, Wiki, or GH discussions.

I'm fine to leave your change to PageNav as it is for this merge, but can we also add the three line comment I suggested to Controller?

samiebee43 (approved):
Thank you for the comment addition.

It highlights the strangeness of the or-statement. In the future it would be nice if doing_raman stayed true whenever it is logically selected in the app, but for now that's OK.

Something like that can be achieved in the future by handling click events something like this
```python
click raman:
    raman = True
    nonraman = False
    expert = False
click non-raman:
    raman = False
    nonraman = True
    expert = False
click expert:
    expert = True
```

PR #279, opened 2023-09-13
Main+4.0.15

samiebee43:
This fast-forwards main to the tag 4.0.15

Technically, this is done by creating a branch from main, merging tag 4.0.15 via CLI, and now opening this pull request between the two branches. Github isn't able to PR a tag otherwise.


samiebee43:
After looking at other PRs, this one should only be used if there's any delta left after merging those. Converting as draft for now.

If there's no delta after--close this.
If there is a delta--this has to be merged before anything 4.0.16 related can be, to preserve the history.

samiebee43:
Unfortunately this has deltas and was made out-of-date by the other PR. That suggests that 4.0.14..4.0.15 and 4.0.14..mzieg-rev4a are mutually exclusive. In other words that means that both 4.0.15 and mzieg-rev4a have commits that the other don't. It also means that the main branch will never be able to coincide with 4.0.15. The closest thing would be to merge this and create a commit on main that has 4.0.15+some stuff.

It's still possible to use the tag 4.0.15, but it's not possible to, for example, git-bisect to this particular version. I would like to highlight two possible next steps.

(1) Merge this, accept the imperfect history -- main will contain a superset of 4.0.15 (it's a subset now, pre-merge)
(2) Rescind the existing 4.0.15 binary and tag, merge this, retag and rebuild from the commit it creates.

Leaning towards (1). What say you @mzieg ?

samiebee43:
Nevermind. I checked the above statements using the CLI and it showed that mzieg-rev4a and 4.0.15 are in fact equivalent.

Turns out Github won't show the latest diff until you update the branch. The best action here is to close this.

mzieg:
Thanks for your diligence in checking this and not taking anyone's (or any system's) word for it :-)

PR #281, opened 2023-09-14
fixed some Mac bits missed in earlier testing

mzieg:
I think before I'd probably built the resources (uic/rcc) from Windows. Works from Mac again.


enlighten/assets/uic_qrc/images/splash.png
samiebee43, line 1: thanks to 'Onion Skin' feature in git I can see that the grey bar moved up slightly. (leaving this comment for future reviewers)

samiebee43 (approved):
Nice

enlighten/assets/uic_qrc/images/splash.png
mzieg: It's also slightly darker, but the important thing is the aliasing was done in Photoshop instead of whatever MSPaint-crap I was pulling in the initial draft :-)

Issue #275, opened 2023-09-12
4.0.14+ sometimes fail to save spectra


samiebee43:

samiebee43:
Not sure what caused this, but it has been addressed by #282.

When someone comes back with an error message, then we reopen the issue.

PR #282, opened 2023-09-14
Sbee fts

samiebee43:
This is how I'm addressing #275

### DIFF
--- a/enlighten/common.py
+++ b/enlighten/common.py
@@ -134,3 +134,44 @@ def get_default_data_dir():
     if os.name == "nt":
         return os.path.join(os.path.expanduser("~"), "Documents", "EnlightenSpectra")
     return os.path.join(os.environ["HOME"], "EnlightenSpectra")
+
+
+"""
+Looking for a method to display Marquee messages? 
+(Those non-intrusive messages at the top of Enlighten)
+
+One way is to use self.ctl.marquee.info("Message")
+This assumes your FeatureObject class has an instance of Controller.
+"""
+
+def msgbox(prompt, title="Alert", buttons=0):
+    """
+    Display an interupting message to the user.
+
+    Inspired by VB msgbox: 
+    https://learn.microsoft.com/en-us/office/vba/language/reference/user-interface-help/msgbox-function
+    """
+
+    # doing this import within this method to avoid breaking all dependents of common.py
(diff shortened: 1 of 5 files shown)


enlighten/common.py
```diff
@@ -134,3 +134,44 @@ def get_default_data_dir():
...
+
+One way is to use self.ctl.marquee.info("Message")
+This assumes your FeatureObject class has an instance of Controller.
+"""
+
+def msgbox(prompt, title="Alert", buttons=0):
+    """
+    Display an interupting message to the user.
+
+    Inspired by VB msgbox:
```
mzieg, line 151: > Inspired by VB

🤮 (but I get it :-)

enlighten/measurement/MeasurementFactory.py
```diff
@@ -88,21 +90,31 @@ def create_from_spectrometer(self, spec, is_collapsed, generate_thumbnail=True,
...
-            save_options = self.save_options,
-            spec         = spec,
-            measurements = self.measurements)
+        try:
+            measurement = Measurement(
+                save_options = self.save_options,
+                spec         = spec,
+                measurements = self.measurements)
+        except:
+            msgbox("Failed to create measurement", "Error")
```
mzieg, line 99: Note that if you add a log.error(exc_info=1) we'll get the stacktrace.

Honestly, I sometimes include stacktraces in on-screen displays...makes it easier for the user to just send me a screenshot.

mzieg (approved):

enlighten/measurement/MeasurementFactory.py
samiebee43: Nice idea. I've implemented this without modifying msgbox, using the traceback module

enlighten/common.py
samiebee43: I'm breaking my own rule of not borrowing culture from other languages, but I remember my first day on Enlighten, I wanted nothing more than to say MSGBOX !

I think it's pythonic enough,,, except the buttons parameter taking an int like the original VB might not be worth keeping around. Maybe we do away with the button parameter and have msgbox always give only an OK button.

Then we can borrow from javascript and use different functions for different alert box behavior: 
msgconfirm() --> Ok/Cancel
msgprompt() --> Yes/No

PR #285, opened 2023-09-15
Add Burn-In plugin

mzieg:
Convenience plug-in for automated hardware load test.

### DIFF
--- a/enlighten/Controller.py
+++ b/enlighten/Controller.py
@@ -296,7 +296,7 @@ class Controller:
         # The background thread only applies changes BETWEEN spectra, so give it
         # time to collect at least one more acquisition so the changes can take 
         # effect (cap at 5sec).
-        final_delay_sec = max(5, float(2 * spec.settings.state.integration_time_ms / 1000.0 + 0.200))
+        final_delay_sec = min(5, float(2 * spec.settings.state.integration_time_ms / 1000.0 + 0.200))
         log.debug("disconnect_device[%s]: final delay of %.2fsec", device_id, final_delay_sec)
         time.sleep(final_delay_sec)
 
--- a/enlighten/Controller.py
+++ b/enlighten/Controller.py
@@ -324,7 +324,7 @@ class Controller:
         return True
 
     def close(self, event_arg_str):
-        log.critical("closing")
+        log.critical(f"closing ({event_arg_str})")
 
         self.shutting_down = True
 
--- a/enlighten/Controller.py
+++ b/enlighten/Controller.py
@@ -2004,11 +2004,11 @@ class Controller:
         # @returns True if graph was updated
         """
         if curve is None:
-            log.error("set_curve_data[%s]: no curve")
+            log.error("set_curve_data[%s]: no curve", label)
             return False
 
         if y is None:
-            log.error("set_curve_data[%s]: no y")
+            log.error("set_curve_data[%s]: no y", label)
             return False
 
         if x is None:
--- a/enlighten/Controller.py
+++ b/enlighten/Controller.py
@@ -258,7 +258,8 @@ class Controller:
         instantiated and displayed with all event loops running.
         """
         log.debug("performing post-construction initialization")
-        self.plugin_controller.autoload(self.autoload_plugin)
+        if self.autoload_plugin:
+            self.plugin_controller.autoload(self.autoload_plugin)
 
     def disconnect_device(self, spec=None, closing=False):
         if spec in self.other_devices:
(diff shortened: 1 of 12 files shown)


mzieg:
Works pretty much as intended in testing.

The laser_expiry was a last-minute add for ambient models.
image

enlighten/Plugins/PluginController.py
```diff
@@ -471,10 +471,19 @@ def connected_callback(self):
...
 
-            # allow the user to enable the plugin
-            self.cb_enabled.setEnabled(True)
-            self.cb_enabled.setChecked(False)
-            self.button_process.setEnabled(True)
+            module_config = self.get_current_configuration()
+            if module_config.auto_enable:
+                # this plugin has requested to auto-enable on connect
+                log.debug("auto-enabling")
+                self.cb_enabled.setEnabled(not module_config.lock_enable)
```
mzieg, line 478: Note plugins can set `self.lock_enable = True` to prevent users from deactivating the plugin once connected.

enlighten/Plugins/PluginController.py
```diff
@@ -1289,6 +1307,14 @@ def apply_commands(self, response):
             log.debug(f"applying {setting}")
             self.multispec.change_device_setting(setting, value)
 
+    def apply_signals(self, response):
+        if response.signals is None:
+            return
+
+        for signal in response.signals:
+            log.debug(f"applying signal: {signal}")
+            eval(signal)
```
mzieg, line 1316: This is awful. I'm sorry. I haven't figured out the "right" way to do it.

scripts/run-burnin.bat
```diff
@@ -0,0 +1,10 @@
+@echo off
+
+:START
+
+python scripts\Enlighten.py --log-level debug --plugin Prod.BurnIn 1>>enlighten.out 2>>enlighten.err
+
+echo Sleeping 10 sec...ctrl-C to cancel
+sleep 10
+
+goto START
```
mzieg, line 10: Having opened with an eval, let's close with a goto. I have no shame :-(

enlighten/Controller.py
```diff
@@ -294,8 +295,8 @@ def disconnect_device(self, spec=None, closing=False):
 
         # The background thread only applies changes BETWEEN spectra, so give it
         # time to collect at least one more acquisition so the changes can take 
-        # effect.
-        final_delay_sec = float(2 * spec.settings.state.integration_time_ms / 1000.0 + 0.200)
+        # effect (cap at 5sec).
+        final_delay_sec = max(5, float(2 * spec.settings.state.integration_time_ms / 1000.0 + 0.200))
```
mzieg, line 299: Ugh, that should be min

mzieg:
merging to main so I can build a distributable installer so I can run several computers in parallel

enlighten/file_io/Configuration.py
```diff
@@ -177,14 +177,14 @@ def load_text(self):
 
     def parse(self):
         """ Load as ConfigParser object. """
-        self.config = configparser.ConfigParser(interpolation=None)
+        self.config = configparser.ConfigParser(interpolation=None, strict=False)
```
mzieg, line 180: This allows enlighten.ini files with duplicate keys to be loaded (don't crash)

enlighten/file_io/Configuration.py
```diff
@@ -256,12 +256,13 @@ def dump_keys():
...
                         # was it a key-value line?
-                        m = re.match(r"^([A-Za-z0-9_]+) *=", line)
+                        m = re.match(r"^([A-Za-z0-9_ ]+) *=", line)
                         if m and section is not None:
                             # it was a key-value line, so update the line with the current value
-                            key = m.group(1)
-                            self.write(outfile, "%s = %s" % (key, self.get(section, key, raw=True)))
-                            seen[section].add(key)
+                            key = m.group(1).strip()
+                            if key not in seen[section]:
```
mzieg, line 263: This ensures we don't re-write duplicate keys we allowed to be loaded from existing files (i.e. cleanup non-conformant files on write)

enlighten/Plugins/PluginController.py
```diff
@@ -499,7 +508,7 @@ def do_post_disconnect(self):
     # plus set the "process" button to the opposite state
     def enabled_callback(self):
         self.enabled = self.cb_enabled.isChecked()
-        if self.button_process is not None:
+        if self.button_process is not None:     # MZ: when would that be none?
```
samiebee43, line 511: Trying to make the habit of using msgboxes to expose things, you could add an else with an `msgbox("unexpected value button_process is not None")`.

The rule for the msgbox's is that the message should be short enough and distinctive enough that someone non-technical can remember it an hour after seeing it.

If two msgboxes are very similar, then I like to vary the wording
"file read error" == "corrupted file" == "fatal error: FILE IO"

We can always go back and prove out a question like "when would this be none", so this is a way to expedite that when we don't have time... (most of the time).

scripts/run-burnin.bat
samiebee43: I recently wrote a C program with goto in it. 
(Not for work)

It was equal parts liberating and buggy.

enlighten/file_io/Configuration.py
samiebee43: I'm liking the inline-github comments. This one should be added to the source file too.

enlighten/file_io/Configuration.py
samiebee43: Add to source

samiebee43: brief. 
```
# ensure no duplicate keys
```

PR #287, opened 2023-09-18
show msgbox warning if max_integ is zero

### LINKED ISSUE #278: Enlighten silently fails to set integration time if max_integration_time=0
image

The best way to handle this is an error dialog "Maximum integration time is set to zero. Please correct your EEPROM." 
Optionally for that session we would use a default of 1000ms.

Importantly, we will not substitute with a default value without notifying the user the EEPROM is incorrect. Would rather fail loudly then quietly do something unexpected.

samiebee43:
resolve #278


mzieg (approved):

mzieg:
adding to tonight's installer

PR #289, opened 2023-09-18
Reference spec.fwhm instead of using default value 3

### LINKED ISSUE #133: RamanShiftCorrectionFeature find_peaks FWHM
The 'width' parameter in RamanShiftCorrectionFeature's call to scipy.signal.find_peaks should probably be tied to Spectrometer.fwhm rather than a hardcoded constant for all spectrometers.  (The current implementation has apparently been found to work on all tested units, but it still seems like a design weakness.)

samiebee43:
Close #133


mzieg (approved):

mzieg:
adding to tonight's installer

PR #293, opened 2023-09-20
BurnIn learnings from using installer

mzieg:
- improved plugin logging so we don't feel compelled to run under DEBUG
- added --log-append so we can aggregate logs if we choose


mzieg:
merging so we can post internal installer tonight

mzieg:
Check Mode buttons...

samiebee43:
> merging so we can post internal installer tonight

This proved to be effective. When I'm not available to review code, it's helpful that main represents the latest installer. When I started working on log file truncation, after a week long break, I was able to build off of your existing append work by simply pulling from main.

Issue #92, opened 2023-01-27
Revisit Marquee animation in ENLIGHTEN 4.x


mzieg:
The old ENLIGHTEN 3.x Marquee animation was cute in that it shrunk the graph, but didn't affect the controls.  It had the flaw that you couldn't SEE the Marquee on any screen other than Scope, which is why we moved it to a higher frame in ENLIGHTEN 4.  However, now the animation actually moves the buttons on...well, every button on the GUI.  And that's not good.  

We need to decide where to put it where it doesn't move UI controls.  I'd like to keep it "animated" (appearing / disappearing as needed), but without disrupting other screen elements.  I'm leaning toward having it unfold in the upper-right corner, where it would overlay the Help and Light Mode buttons (same place MacOS notifications appear).

samiebee43:
Here is a screenshot that shows the moving buttons.

![i92-original](https://user-images.githubusercontent.com/124081765/216396032-c17385e7-ce42-4dfd-97cb-d15bef4b602d.png)

One solution might be to leave a dedicated space in the UI for the notification bar.

![i92-dedicated](https://user-images.githubusercontent.com/124081765/216395882-43af4cef-695e-4608-ac2e-902c33a66f72.png)

Another option is to squish the nav bar.

![i92-squashlogo](https://user-images.githubusercontent.com/124081765/216395891-2e088239-fd6e-46b3-bc95-93ada60a5e8a.png)

Finally, there is the floating notification option. Which I think is probably the best one. It has the added benefit that the width can be reduced to just cover the text needed.

![i92-floating](https://user-images.githubusercontent.com/124081765/216398929-e0bde55e-0fae-49fd-9fb5-4f93e63ce5a3.png)

mzieg:
Hi Samie, for now I vote for "dedicated space in the UI for the notification bar."  I conceptually like the "floating" alert, but it covers up other buttons.  To me, your first suggestion solves the immediate customer-facing problem while buying us time to play with other display options in-house.

samiebee43:
Pretty sure this is fixed now. Need to confirm if it has been merged to main

samiebee43:
Dedicated space is merged to main and available in 4.0.12 beta.

PR #296, opened 2023-09-25
Set Enlighten to default to truncated appending logs

samiebee43:
Using `--log-append=LIMIT` as the new default, Enlighten will append to the log file but keep it within a certain size.

The size limit is decided by Wasatch.PY. Currently it is 20mb, but we want to tweak that based on a few things.

1) start up time. At 20mb it takes more than 60 seconds to open Enlighten. We want to keep launch sub, say, 35 seconds.
2) log coverage time. I'm aiming for 3-10 hours of program time if it's possible.
3) email attachment size. With a max of 25mb on most email clients, that's where 20 mb comes from (~5mb for last session, which may not have a program start truncation)

### DIFF
--- a/scripts/Enlighten.py
+++ b/scripts/Enlighten.py
@@ -52,7 +52,10 @@ class EnlightenApplication(object):
 
     ## Parse command-line arguments
     def parse_args(self, argv):
+
+        # SB: this doesn't show up anywhere bc logging is not yet configured
         log.debug("Process args: %s", argv)
+
         self.args = self.parser.parse_args(argv)
         if self.args is None:
             return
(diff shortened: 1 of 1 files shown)


samiebee43:
We ended up with about 7 minutes of program time for 20mb not 3-10 hours :(

start up time around 20mb is garbage, but it's not because of the file trimming code. That executed on a 60gb file instantly. (Good job, me). More info here: https://github.com/WasatchPhotonics/Wasatch.PY/pull/89

samiebee43:
We may want to control Enlighten's boot time when the log is in the ballpark of 20mb. It's currently 1m 30s.

Thanks to the gray bubble in the splash screen, we can see that ResourceMonitorFeature _seems_ to be the thing lagging in that case, but I don't see anything in that file that would reference the log. Notably, ResourceMonitorFeature is the first of "create_rest" in BusinessObjects

scripts/Enlighten.py
```diff
@@ -78,16 +81,21 @@ def parse_args(self, argv):
...
-        parser.add_argument("--stylesheet-path",    type=str,                       help="Path to CSS directory")
-        parser.add_argument("--headless",           action="store_true",            help="Run Enlighten without GUI")
-        parser.add_argument("--plugin",             type=str,                       help="Plugin name")
+
+        # This code was like a spreadsheet when I found it, leaning into that
+        # use :set nowrap in vim
+        # TODO: everything should have a default -- do not rely on empty args
+        #                  | parameter name       | type    | default             | action             | choices                                                  | help |
+        parser.add_argument("--log-level",         type=str, default="info",                            choices=['debug', 'info', 'warning', 'error', 'critical'], help="logging level")
+        parser.add_argument("--log-append",        type=str, default="LIMIT",                      choices=["False", "True", "LIMIT"],                   help="append to existing logfile")
```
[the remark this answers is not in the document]
samiebee43: Not quite, no.

Here's how it could go:

## Scenario A, arbitrarily long sessions

1) Using my existing Enlighten with "--log-append=false" by default, I may create some arbitrarily large log file. Enlighten always cleared these files _only at start up time_, so as long as the most recent one was a fairly log-heavy long session, the file may get quite large.

2) Using a newly installed Enlighten with "--log-append=limit", the log will be instantly[^1] trimmed to the limit size driven by Wasatch.PY (currently 20mb). Like previous versions, the log can still grow to an arbitrary size if there is a long log-heavy session. Trimming happens only at start up time.

3) The next start up will do the same thing, it will trim, but the left over 20mb seems to slow down the startup.

4) This will happen indefinitely, until the user manually deletes the log file, because once it reaches a 20mb limit, it will always be at least that big.

## Scenario B, several normal sessions that add up

1) Since the log is appending, even if the sessions are individually small they will add up.

2) Eventually the log file will reach capacity and all subsequent start ups will be slowed down.

## Conclusion

The slow boot would eventually affect everyone. My working theory is that it's the 20mb that is causing a slowdown somewhere, and that any size beyond that does not introduce any additional slowdown. That theory is supported by the fact that I observed the same slow start up on Enlighten main and Wasatch on 501e78a3. Still it would not hurt to further unit test the File Resize Program, just in case, I didn't intend for sizes 10gb+ when I wrote it.

The slow boot should be fixed one way or another before merging. So this is now a draft.

## Action Step

Either 

A) We reduce the limit size such that the boot time is acceptable. For example, if 10 sec is okay, then we can figure out the corresponding size (approx 2mb). This is super easy, but it kind of blows--2mb is like 42 seconds of historical logs.

B) (preferred) We eliminate the cause of slowdown pre-merge.

One hacky way to do it would be to keep the enlighten.log from the previous session in enlighten.log.temp until Enlighten is done booting up, then moving it in-place. I don't recommend we actually do that, but thinking that way may shed some light towards an actual solution.

[^1]: It looked like it was able to resize a 50gb file instantly. This can be verified by further unit testing of the [File resize program](https://github.com/WasatchPhotonics/Wasatch.PY/blob/f2a28fc07442ae9c6c3ebb21883381d2d68f8019/wasatch/utils.py#L596).

samiebee43:
Actually, since I don't have time to work on this and BLE at the same time, maybe we should just set the limit to 2mb for now and come back to it later ? 

At least it will help with the case of "let me restart Enlighten real quick before sending the log to support", and it will provide us some surface to extend our log history later.

samiebee43:
Now [Wasatch.PY #92](https://github.com/WasatchPhotonics/Wasatch.PY/pull/92) is a prerequisite.

scripts/Enlighten.py
```diff
@@ -77,18 +80,26 @@ def parse_args(self, argv):
     ## Defines the command-line arguments and their defaults
     #
     def create_parser(self):
+<<<<<<< HEAD
```
mzieg, line 83: Assuming we just need to delete this?

samiebee43: my bad.. thx for handling

PR #297, opened 2023-09-26
Remove disable laser button when battery low

samiebee43:
This feature was causing issues on a unit plugged to power, removing.

It also had a bug in the tooltip which makes me think the low-battery laser disable was never commonly used.
That is an example of over-engineering, btw.


samiebee43:
We agreed offline to set the EEPROM value `has_battery` to zero instead.

PR #300, opened 2023-09-28
Small ENLIGHTEN tweaks

mzieg:
- sleep isn't a DOS command
- replace --headless with --window-state choices

### DIFF
--- a/enlighten/Controller.py
+++ b/enlighten/Controller.py
@@ -97,7 +97,7 @@ class Controller:
                 set_all_dfu       = False,
                 form              = None,
                 splash            = None,
-                headless          = False,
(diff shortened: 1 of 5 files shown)


mzieg:
Note difference between "maximized" (top) and "fullscreen" (below). 

sample-maximized

sample-fullscreen

Note there's no [x] in fullscreen, but alt-F4 works.

samiebee43:
https://github.com/WasatchPhotonics/ENLIGHTEN/blob/0ebbfde447098744cabc4ae0768123e6c17f55ab/test/conftest.py#L30

change to  --window-state minimized

PR #308, opened 2023-10-06
Windows installer libusb0

mzieg:
Something broke when tensorflow forced some dependency updates, because now we're getting the old "libusb backend not found" when run from the compiled Windows installer, even though it works fine from source.  Testing possible fixes...

### DIFF
--- a/README_CHANGELOG.md
+++ b/README_CHANGELOG.md
@@ -2,6 +2,8 @@
 
 # 4.0 GUI and Plugin Refactoring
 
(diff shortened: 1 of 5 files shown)


mzieg:
Merging so I can start another branch for testing.  I probably should have created a 4.1.0-dev branch earlier, so that all of these "release-prep" and "test-fix" branches could have been off that, and the RC installers built from that, and then we could merge everything up to main when 4.1.0 was ready to be cut. Sorry.

PR #310, opened 2023-10-07
Fixed --plugin autoload

mzieg:
The --plugin command-line had been working from source, and "sort-of" worked from the installer...depending on what calls you make from the plug-in.  I haven't entirely worked out what the rules are (and am too tired to think it through), but it "felt right" that delaying plugin initialization until after the spectrometer was fully connected was likely to simplify all manner of potential race conditions. And it seemed to fix things in testing (from the installer).

- delayed --plugin autoload until after first spectrometer connects
- add version number to --help

Also, cut a 4.1.0-dev branch :-)


mzieg:
It's possible this broke something in another project.  Just tagging for reference.

samiebee43:
Is this ready for review?

mzieg:
I thought it was, but I think I need to do some more testing myself.  Moved back to draft.

mzieg:
OBE

PR #311, opened 2023-10-07
Deprecate Plugin Dependencies

mzieg:
- support 'events' in functional plugins
- remove plugin 'dependencies' (not needed with access to the Controller)
- hide plugin field frame if no fields

### DIFF
--- a/README_CHANGELOG.md
+++ b/README_CHANGELOG.md
@@ -4,6 +4,11 @@
 
 ## 4.1.0-Dev
 
+- 2023-??-?? 4.0.25
+    - plugins
+        - hide plugin field frame if no fields
+        - deprecate dependencies
+        - support events in functional plugins
 - 2023-10-09 4.0.24
     - Prod.BurnIn temperature fix
     - added run_from_root
--- a/enlighten/Plugins/PluginController.py
+++ b/enlighten/Plugins/PluginController.py
@@ -455,9 +455,6 @@ class PluginController:
             #
             ####################################################################
 
-            # This is AFTER configure_gui_for_module() is called, meaning 
-            # enlighten_info.dependencies should be populated
-
             connected_ok = False
             try:
                 connected_ok = self.run_worker()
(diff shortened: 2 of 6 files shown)


samiebee43:
Overall I really like the direction this is going. I'm glad we can deprecate plugin_dependencies. 

The only thing I disagree with is the inclusion of an event queue. I rather we dispatch events as soon as we are aware of them, implying a user-space requirement that event handlers take no longer than 17ms to return.

mzieg:
> The only thing I disagree with is the inclusion of an event queue. I rather we dispatch events as soon as we are aware of them, implying a user-space requirement that event handlers take no longer than 17ms to return.

The only real queued in the plugin architecture is the out-going EnlightenPluginResponse.signals array (basically used by Prod.BurnIn) which I agree is clunky...I just haven't figured out the right way to dispatch closures to the Qt GUI thread from inside PluginWorker's threading.Thread.

EnlightenPluginConfiguration.events isn't really a queue, it's meant to hold callback methods when various things happen inside ENLIGHTEN that a plugin might need to know about.

Now that I think about it, since plugins now have access to the Controller and all BusinessObjects, they can simply "register" for the events they care about directly, and we no longer need separate handling of those.

I'll take a look at deprecating events next, in a separate branch.

Note to self: it looks like EnlightenPluginConfiguration.events was a dict, but EnlightenPluginBase._events is a list...confusing.

samiebee43:
I wrote about how I think plugins should deal with events in the future here: https://github.com/WasatchPhotonics/ENLIGHTEN/wiki/EFPlugins#user-concept

The main idea is that the user implements functions with special names based on the event they handle. The plugin system has all of those names hardcoded. When an event happens, say on_gui, the plugin system will check if the plugin has implemented such a function, and if so it'll call it immediately.

mzieg:
Hi Samie, any other changes required for this to go into 4.1.0?  The ability to use PluginController's existing event callbacks from functional plugins will be useful.

samiebee43 (approved):

PR #312, opened 2023-10-09
Chdir() at launch

### LINKED ISSUE #294: asset relative paths
When running 4.0.17's run-burnin.bat from the Windows installer I found errors about missing ASTM JSON file.  That's probably because it's loaded from a relative path enlighten/assets/example_data/... and run-burnin.bat is run from plugins/Prod.  Consider doing an internal "change directory" as soon as ENLIGHTEN launches so it's running from the directory containing the .exe, or the directory "above" Enlighten.py if run from source.

mzieg:
- suggested by @samiebee43 to resolve #294 
- also trying to find a way to keep the console window suppressed on Win10 and Win11 without --windowed (which messes up other things) (see https://github.com/pyinstaller/pyinstaller/pull/7735)

### DIFF
--- a/README_CHANGELOG.md
+++ b/README_CHANGELOG.md
@@ -4,9 +4,12 @@
 
 ## 4.1.0-Dev
 
-- 2023-10-09 4.0.23
+- 2023-10-09 4.0.24
     - added run_from_root
-    - changed --hide-console to hide-late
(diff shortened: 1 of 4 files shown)


mzieg:
Tested on Win11 laptop via installer. Chdir() works as expected, console window at least minimizes (doesn't appear at all on Win10).

samiebee43 (approved):

PR #314, opened 2023-10-11
resolve #183 invalid default string next to Save .ini File btn

### LINKED ISSUE #183: Invalid path shown next to Save .ini button

samiebee43:
#183 is labeled as closed, so why does this old branch appear ahead of main by one commit ?

We can even see the line this introduces on main:
https://github.com/WasatchPhotonics/ENLIGHTEN/blob/82d913ce2497e89d68243dc057d81d8c62c203e6/enlighten/file_io/Configuration.py#L92

### DIFF
--- a/enlighten/file_io/Configuration.py
+++ b/enlighten/file_io/Configuration.py
@@ -14,7 +14,7 @@ from enlighten import common
 
 log = logging.getLogger(__name__)
 
-class Configuration(object):
+class Configuration:
     """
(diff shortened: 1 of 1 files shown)


samiebee43:
Also, this means sbee-doc is safe for deletion and should be deleted.

PR #316, opened 2023-10-11
Remove FWHM from Local Baseline Plugin on main

samiebee43:
Overall I want to retire my FWHM program. 

It's not well written, and it strains the already barely working plugin/annotation system. The approach is okay, and the in-line diagrams are nice in my opinion, but the use of relative indices (continuing from Local Baseline's use of the same) makes it bug prone and difficult to maintain.

Nobody is depending on this program afaik, and I would like to remove it before that changes. It's due for a rewrite--best kept separate from Local Baseline (because of the relative indices problem).


mzieg:
I'm okay with removing the FWHM functionality from it, but I'd leave the LocalBaseline.py file in the pluginExamples/Analysis directory.  It looks like it got deleted and re-added at the root?

samiebee43:
Thanks for catching that !

Issue #240, opened 2023-07-12
Apply the "persistent variables style guide"


samiebee43:
The "persistent variables style guide" is as follows.

## Introduction

I would like to add some formality to how we deal with persistent variables, such as GainDB. The objective is to simplify and beautify existing code, make it easier to enable/disable persistence for new variables, and to standardize the implementation specifics (such as order of operations) across all variables. 

During the lifetime of an Enlighten instance, such variables have three major life events: initialization, modification, and release.

### initialization
(1) check the .ini path 
(2) check memory (SpectrometerSettings, in case of multispec)
(3) check the eeprom
(4) use a hardcoded default

The search path should appear sequentially in code. Pseudo:
```python
# do something like this
def __init__(self):
    # initialize to None if necessary to silence a static analyzer
    self.gain_db = None
    if ini_available() and ini_has_path(sn, "gain_db"):
        self.gain_db = get_from_ini(sn, "gain_db")
    elif eeprom_has_value("gain_db_default"):
        self.gain_db = get_from_eeprom("gain_db_default")
    else:
        self.gain_db = 8

# do NOT do this
# this is a 'define and replace' pattern, often useful but not to be used to resolve persistent variables
def __init__(self):
    self.gain_db = 8
    if  eeprom_has_value("gain_db_default"):
        self.gain_db = get_from_eeprom("gain_db_default")
    if ini_available() and ini_has_path(sn, "gain_db"):
        self.gain_db = get_from_ini(sn, "gain_db")

# do NOT do this either
def __init__(self):
    self.gain_db = None
    # code (ini, hardcoded, eeprom) does not match search path (ini, eeprom, hardcoded)
    self.gain_db = get_from_ini(sn, "gain_db", default=8)
    if eeprom_has_value("gain_db_default"):
        self.gain_db = get_from_eeprom("gain_db_default")
```

### modification
(4) set scope-level variable `x = ...`
(5) set class-level variable  `self.x = ...`
(6) set application-level variable `self.ctl.multispec.set_state('x', ...)`
(7) set memory map of ini `self.ctl.config.set('x', ...)`
(8) propagate message to device driver `self.ctl.multispec.change_device_setting('x', ...)`
(9) propagate message to UI ` ctl.form.ui...setValue(...)`*

*This often needs to be wrapped by 'blockSignals' calls. `_quiet_set` should be moved a common UI helper module.

### release
(10) write modifications to ini. This happens when Enlighten is closed normally.

[...]

PR #328, opened 2023-10-17
Sb persist sg

samiebee43:
## What has changed

Adjustments to be more compliant with style guide #240 

Gain, Integration Time, and Boxcar had the following change of pattern for updating persistent parameters. The result is one less state-to-state copying.
image

Gain had a rewrite of its load from persistence such that the code is structured to match the searchpath specified in the issue above.

## Todo

- ~Test persistence of these three variables on XS~
- ~scan for references of state.gain, etc~

### DIFF
--- a/enlighten/BusinessObjects.py
+++ b/enlighten/BusinessObjects.py
@@ -607,6 +607,7 @@ class BusinessObjects:
 
         self.header("instantiating BoxcarFeature")
         ctl.boxcar = BoxcarFeature(
+            ctl,
             bt_dn                       = sfu.pushButton_boxcar_half_width_dn,
             bt_up                       = sfu.pushButton_boxcar_half_width_up,
             spinbox                     = sfu.spinBox_boxcar_half_width,
@@ -617,6 +618,7 @@ class BusinessObjects:
 
         self.header("instantiating IntegrationTimeFeature")
         ctl.integration_time_feature = IntegrationTimeFeature(
+            ctl,
             bt_dn                       = sfu.pushButton_integration_time_ms_dn,
             bt_up                       = sfu.pushButton_integration_time_ms_up,
             marquee                     = ctl.marquee,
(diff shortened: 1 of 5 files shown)


samiebee43:
This is a significant portion of the diff that I'd like to highlight. No long in Configuration's save_file routine do we last-minute copy from one memory location to another. This also has the clue that state.integration_time_ms et al will need to have all of its references replaced with `self.ctl.config.get(...)` (thus config can be used for inter-session and same-session persistence)

Configuration.py 
```diff
...
def save_file(...):
    ...
-    self.set(sn, "integration_time_ms", state.integration_time_ms)
-    self.set(sn, "boxcar_half_width", state.boxcar_half_width)
-    self.set(sn, "gain_db", state.gain_db)
    ...
```

samiebee43:
## Scan for state references

Unfortunately there are great number of references to state.integration_time_ms et al. It's enough to oversize the scope of this pull request. So instead I put back (coined "re-add") the saves to multispec state. Still we want to change config's memory model immediately instead of waiting for the shutdown event.

Now the approach for persistent variable modification is to write out each stage and clearly comment them:

```python
# save integration time to application state
self.multispec.set_state("integration_time_ms", now_ms)

# persist integration time in .ini
self.ctl.config.set(self.ctl.multispec.current_spectrometer().settings.eeprom.serial_number, "integration_time_ms", ms)

# send integration time change to hardware
spec.change_device_setting("integration_time_ms", now_ms)
```

samiebee43:
> I haven't digested all of this, but if you remove the calls to Multispec, are you eliminating the feature where multiple spectrometers could be controlled at once?

No. Nothing changes in terms of functionality.

mzieg (approved):
This all looks okay. I tested with two connected spectrometers and confirmed multispec.locked still worked, which was the thing I was concerned about.

samiebee43:
This closes the basic requirements of 240, but we should copy it's contents to Discussion or Wiki since it can be used for broader improvements to persistent variables' lifecycle--can be applied to anything that uses `self.ctl.config` `get` or `set`.

(close #240)

PR #334, opened 2023-10-30
Fix graph for XS

samiebee43:
Graph still referenced a non-existent function. When that is updated, the XS is able to show spectra on this branch again.

Once this is merged to mzeig-roi-handles, we'll need to re-review it before merging to 4.1.0-dev


mzieg (approved):

mzieg:
You beat me to it, but today I was processing my mail from bottom-up...

PR #351, opened 2023-11-16
RamanIntensityCorrection less confusing docstring

samiebee43:
The description of RamanIntensityCorrection said that it is not the same as RamanShiftCorrection, but then it defines RamanShiftCorrection before defining what this file is actually about: RamanIntensityCorrection.

Instead, I'll leave the note that this is not RamanShiftCorrection--and I will go on to define only RamanIntensityCorrection. If the reader wants a definition of the conflated term, they can visit that file.


WasatchEngineering (approved):

samiebee43:
Was this approved by @mzieg ?

mzieg (approved):
My alter-identify _also_ approves these changes.

PR #354, opened 2023-11-27
4.1.0 arm dist

samiebee43:


enlighten/common.py
```diff
@@ -13,7 +13,7 @@
       can be modules (files) within it
 """
 
-VERSION = "4.0.26"
```
mzieg, line 16: I'd rather not jump to 4.1.0 with this commit.  The first 4.1.0 release should be for Windows and on our website, not a Mac RC.

mzieg: Hi Samie, I think I'd rather go with my Mac branch instead of this one. I don't want to start committing the generated Python files.

samiebee43: Sounds good. I'll close this one

Issue #396, opened 2024-05-07
Laser not allowing immediate manual turn-off in Batch Collection


mzieg:
"I’ve been noticing an issue with Enlighten 4.0.58 where the laser takes a long time (up to ~30 seconds) to turn off after I click the “turn laser off”  I’m a bit worried about this causing a safety hazard.  It only seems to happen when I am in batch mode.  If I am taking 30 second integrations (15 samples at 2000 ms each) and I stop the batch early and click the laser off button, it will take ~30 seconds to turn off the laser."

mzieg:
I think the fundamental request was fixed by a quick change to BatchCollection.stop, but using this to explore an other reported issue in BatchCollection with LaserMode.Spectrum. I'd thought that was isolated to XS, but it seems to be affecting X/XM as well :-(

The problem is highlighted in the below; the first TakeOneRequest (with the laser enable flags raised) is correct and valid; the second is the problem. Digging...

```
2024-05-07 17:53:00,719 [0x00002108] enlighten.timing.BatchCollection DEBUG    BatchCollection.start_collection: instantiating TakeOneRequest in laser_mode.spectrum
2024-05-07 17:53:00,719 [0x00002108] wasatch.TakeOneRequest DEBUG    instantiated TakeOneRequest <id 03916-08456-1715118780719668400-24385, avg 1, take_dark False, enable_laser_before True, disable_laser_after True, laser_warmup_ms 3000>
2024-05-07 17:53:00,719 [0x00002108] wasatch.TakeOneRequest DEBUG    --> via: File "Y:\code\ENLIGHTEN\scripts\Enlighten.py", line 262, in <module>
2024-05-07 17:53:00,719 [0x00002108] wasatch.TakeOneRequest DEBUG    --> via: File "Y:\code\ENLIGHTEN\scripts\Enlighten.py", line 248, in main
2024-05-07 17:53:00,719 [0x00002108] wasatch.TakeOneRequest DEBUG    --> via: File "Y:\code\ENLIGHTEN\scripts\Enlighten.py", line 166, in run
2024-05-07 17:53:00,719 [0x00002108] wasatch.TakeOneRequest DEBUG    --> via: File "Y:\code\ENLIGHTEN\enlighten\ui\VCRControls.py", line 246, in start_collection
[...]

PR #398, opened 2024-05-07
Fixing BatchCollection with LaserState.Spectrum

mzieg:
- fixes long-standing issue (a few months) where LaserState.Spectrum had stopped working, probably due to some changes made for XS
- also addresses #396

### DIFF
--- a/enlighten/timing/BatchCollection.py
+++ b/enlighten/timing/BatchCollection.py
@@ -348,15 +348,10 @@ class BatchCollection:
 
             # initialize driver-level laser auto_triggering and dark collection
             if self.laser_mode == "spectrum":
-                log.debug("BatchCollection.start_collection: instantiating template TakeOneRequest with laser_mode.spectrum")
                 self.take_one_template = TakeOneRequest(take_dark=self.dark_before_batch, enable_laser_before=True, disable_laser_after=True, laser_warmup_ms=self.laser_warmup_ms)
             else:
-                log.debug("BatchCollection.start_collection: instantiating template TakeOneRequest (default)")
                 self.take_one_template = TakeOneRequest()
 
-            # log.debug(f"starting laser_mode {self.laser_mode}, so take_one_request {take_one_request}")
-            # self.ctl.multispec.change_device_setting("take_one_request", take_one_request)
-            # self.ctl.multispec.set_app_state("take_one_request", take_one_request)
             self.running = True
 
         else:
--- a/enlighten/timing/BatchCollection.py
+++ b/enlighten/timing/BatchCollection.py
@@ -176,6 +176,7 @@ class BatchCollection:
         self.clear_before_batch = False
         self.export_after_batch = False
         self.collection_start_time = None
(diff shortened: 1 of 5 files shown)


enlighten/timing/BatchCollection.py
```diff
@@ -345,13 +348,10 @@ def start_collection(self):
 
             # initialize driver-level laser auto_triggering and dark collection
             if self.laser_mode == "spectrum":
-                take_one_request = TakeOneRequest(take_dark=self.dark_before_batch, enable_laser_before=True, disable_laser_after=True, laser_warmup_ms=self.laser_warmup_ms)
+                self.take_one_template = TakeOneRequest(take_dark=self.dark_before_batch, enable_laser_before=True, disable_laser_after=True, laser_warmup_ms=self.laser_warmup_ms)
             else:
-                take_one_request = TakeOneRequest()
+                self.take_one_template = TakeOneRequest()
```
mzieg, line 353: This fixed the issue with "spectrum" laser_mode no longer working. Before, BatchCollection had been "pushing" a TakeOneRequest with the correct enable_laser_before and disable_laser_after fields in start_collection, but it was assuming that TakeOneRequest would then "persist" and be used throughout the batch. 

That's not how it works. BatchCollection works by virtually clicking the "Step-and-Save" (⏭️) button of VCRControls at timed intervals. VCRControls.step_and_save in turn calls TakeOneFeature.start, which generates its own TakeOneRequest to feed downstream into Wasatch.PY.

(There was a time when BatchCollection would "persist" settings like "auto_laser_enable" inside Wasatch.PY, but it got ugly and didn't scale, so we recently added TakeOneRequest so that all requested settings for a given measurement would be encapsulated in a single object, with a unique ID, so we could then "catch" the resulting spectrum and associate it, by ID, with the originating request.)

So anyway, my solution was for BatchCollection to define a "template" of how it wanted TakeOneRequests to be generated for the purpose of the batch, and then pass that template into VCRControls when triggering measurements.

enlighten/timing/BatchCollection.py
```diff
@@ -499,6 +499,11 @@ def stop(self):
         self.running = False
         self.timer_measurement.stop()
         self.timer_batch.stop()
+
+        if self.laser_mode != "manual":
+            log.debug(f"BatchCollection.stop: disabling all lasers (laser_mode {self.laser_mode})")
+            self.ctl.laser_control.set_laser_enable(False, all_=True)
+
```
mzieg, line 506: This fixed the issue where people were hitting "Stop" (⏹️) during a BatchCollection, and it wasn't turning off the laser, even if the laser had been automatically enabled as part of the BatchCollection settings.

jkleinwp (approved):
Not exactly clear on how to reproduce the error so I didn't specifically test that case.

However, changes look good to me and the program is nominally working with batch data collection enabled / spectrum laser mode.

Platform: macOS 14.3.1, WP-01519, 1.0.16.3 MC FW, 01.4.13 FPGA FW

PR #401, opened 2024-05-09
Prompt user to confirm overwrite of existing file

### LINKED ISSUE #392: Prompt to confirm before overwriting existing file on export
This doesn't happen with the "default" filenames (which have timestamps), but if the user manually-enters a filename, they may accidentally overwrite old data.

jkleinwp:
Check for existing filename with .csv or .json extension and prompt the user to overwrite (or not) if detected.

### DIFF
--- a/enlighten/measurement/Measurements.py
+++ b/enlighten/measurement/Measurements.py
@@ -386,20 +386,22 @@ class Measurements:
         directory = common.get_default_data_dir()
 
         # warn user if they are about to overwrite an existing file
-        file_exists = os.path.exists(os.path.join(directory, filename))
+        # it looks like only risk is for .csv and .json extensions
+        file_exists = (os.path.exists(os.path.join(directory, f'{filename}.csv'))
+                       or os.path.exists(os.path.join(directory, f'{filename}.json')))
 
-        log.info(f"file_exists checking for: {os.path.join(directory, filename)}, result={file_exists}")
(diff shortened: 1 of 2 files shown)


enlighten/measurement/Measurements.py
```diff
@@ -385,19 +385,34 @@ def export_session(self, filename=None, prompt=True):
         # currently, all Sessions are stored in ~/EnlightenSpectra
         directory = common.get_default_data_dir()
 
-        # doesn't use the dict
-        if self.ctl.save_options.save_csv():
-            self.export_session_csv(directory, filename, visible_only=visible_only)
+        # warn user if they are about to overwrite an existing file
+        # it looks like only risk is for .csv and .json extensions
+        file_exists = (os.path.exists(os.path.join(directory, f'{filename}.csv'))
+                       or os.path.exists(os.path.join(directory, f'{filename}.json')))
```
mzieg, line 391: Technically, I suppose .spc could go in there too. I don't think many people export to .spc but _shrug._

mzieg (approved):

enlighten/measurement/Measurements.py
jkleinwp: Got it.  Will change. I had all the export boxes checked, but I did't have the .spc (or jcamp) writer installed correctly so it wasn't exporting.

PR #402, opened 2024-05-10
display empirical indication of laser status

mzieg:
When firing the laser, grab a copy of the current ambient spectrum, and compare subsequent spectra against that to infer when the laser appears to actually begin firing.

(Branch name: I was going to use Pearson for this, but then decided area-under-the-curve was both simpler and more robust.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
         - always disable laser in BatchCollection.stop
         - added TakeOneFeature copy-ctor (fixed LaserMode.Spectrum)
     - plugins
+        - fixed bug in base class
         - moved EmissionLines and RamanLines to combobox
     - XS
         - added microcontroller serial number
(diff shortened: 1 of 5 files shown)


mzieg:
I might revisit this later and change the laser buttons from red to the new "SRM Orange" when the laser is in the "initializing state" (and perhaps the bottom-right StatusIndicators as well). For today, the Marquee message is enough.

jkleinwp:
I tested the code under nominal conditions and it works as expected.  

However, under certain edge cases, it doesn't work as expected.  For example:
1. Start laser with magnet interlock removed and sample close to the focusing lens
2. Remove sample (signal increases)
3. Logs report laser is believed to be firing / status changes to "laser is initializing" (unclear what initializing is means)

This is an artificial case, but I could image a scenario where the laser is started w/ interlock activated and someone opens a sample port and sw thinks laser is firing.  

I'm not sure if magnet interlock is status is knowable by the SW.

mzieg:
> I'm not sure if magnet interlock is status is knowable by the SW.

The released XS firmware will indeed share "unambiguous" laser firing state with the host. However, the current patch is provided for a customer who is unable to update their firmware and explicitly requested a "software-only" solution. In their particular case, there is no possibility of changing the sample working distance, and there are minimal concerns about ambient light.

So you're right, this is definitely not "the final solution," but it's better that what we had before and meets the need for today.

jkleinwp (approved):

PR #417, opened 2024-05-30
add default FWHM

### LINKED ISSUE #416: RamanShiftCorrection fails if no FWHM configured in EEPROM
For now, provide better default for old units that didn't receive this calibration in WPSC.

mzieg:
- resolves #416

### DIFF
--- a/enlighten/data/ModelFWHM.py
+++ b/enlighten/data/ModelFWHM.py
@@ -54,6 +54,8 @@ TABLE = """
     WP-633-ER-R-50      11.8       11.8      13.5       13.5
 """
 
+DEFAULT = 15
+
 class ModelFWHM:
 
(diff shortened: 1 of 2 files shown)


jkleinwp (approved):

enlighten/data/ModelFWHM.py
```diff
@@ -54,6 +54,8 @@
     WP-633-ER-R-50      11.8       11.8      13.5       13.5
 """
 
+DEFAULT = 15
```
mzieg, line 57: Now that I read this, it should probably be like 15 for Raman instruments (have an excitation), and 3 for non-Raman, since FWHM is a raw scalar and needs interpreted as either 1/cm or nm. But it's still better than what we had so moving forward.

PR #422, opened 2024-06-06
Support plugins using pyserial

jkleinwp:
Added pyserial dependency to support plugins that require serial connection


jkleinwp:
> Have you built a binary Windows installer and confirmed the plugin works from that?

Yes

mzieg (approved):

Issue #450, opened 2024-07-02
Better visual indication in ENLIGHTEN when laser is in LaserWarningDelay state


mzieg:

mzieg:
resolved in 4.0.67

PR #451, opened 2024-07-02
[#450] visual indication of is_laser_firing

mzieg:
It is almost disturbing that Issue 450 should land on pull request 451.


mzieg:
https://www.youtube.com/watch?v=iwuHLN6BrzM

Issue #460, opened 2024-07-25
Can't load just-saved Raman measurement :-(


mzieg:

mzieg:
[ps-20240725-182206-720227-WP-01630.csv](https://github.com/user-attachments/files/16384213/ps-20240725-182206-720227-WP-01630.csv)

mzieg:
bug inadvertently introduced in #452 :-(

mzieg:
resolved in #461

Issue #452, opened 2024-07-11
Incorrectly prompts for overwrite when re-labeling thumbnails


mzieg:
Clearly I introduced a bug in #428 :-(

mzieg:
resolved in 4.0.67

PR #461, opened 2024-07-25
fix issues loading saved spectra

mzieg:
- fix logging bug in MeasurementFactory
- fixed missed calls to add_renamable from #452 
- fix bug graphing loaded .json measurements

### DIFF
--- a/enlighten/measurement/MeasurementFactory.py
+++ b/enlighten/measurement/MeasurementFactory.py
@@ -243,8 +243,8 @@ class MeasurementFactory:
             elif pathname.lower().endswith(".spc"):
                 measurements = self.create_from_spc_file(pathname)
         except:
-            msgbox("failed to parse file %s" % pathname)
-            log.error("failed to parse file %s", pathname, exc_info=1)
+            log.error(f"failed to parse file {pathname}", exc_info=1)
+            msgbox(f"failed to parse file {pathname}")
 
         if measurements is None:
             return
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@
 - 2024-??-?? 4.0.68
     - change StatusBar frame count from ProcessedReading to Reading.session_count
     - clarified BatchCollection "explain this" re: collection timeout
+    - [#460] fix issues loading saved spectra
+        - fix logging bug in MeasurementFactory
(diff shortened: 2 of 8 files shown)


mzieg:
This will resolve #460

enlighten/measurement/MeasurementFactory.py
```diff
@@ -243,8 +243,8 @@ def create_from_file(self, pathname, is_collapsed=False, generate_thumbnail=True
             elif pathname.lower().endswith(".spc"):
                 measurements = self.create_from_spc_file(pathname)
         except:
-            msgbox("failed to parse file %s" % pathname)
```
jkleinwp, line 246: Any reason for the switcharoo?

jkleinwp (approved):

enlighten/measurement/MeasurementFactory.py
mzieg: I wasn't getting the exception stack trace in the log, and I think it's because pathname was being passed to log.error as a positional argument rather than string interpolation. I switched the line order because in general I figure it's important to log the exception before making additional calls to the GUI (msgbox), as that's just another opportunity for something to mess up before you capture the root cause. (In this case I think the real root cause was the "session count" metadata on a loaded measurement which had no original wasatch.Reading object.)

In general our measurement save/load architecture is very loose. When we save a file, we do NOT save "everything" about it (even in JSON), and when we re-load it, we definitely do not end up with an in-memory object which matches the original. This was somewhat deliberate, because when we "save" measurements we're saving "customer-usable representations" which are "sufficient for customer purposes," and they were never really intended to be completely reconstructable source objects. Literally, the "Load" button in ENLIGHTEN had one original purpose: to be able to SHOW A GRAPH TRACE on the scope for comparison against "live" data. There's a whole bunch of stuff in a Measurement object that you don't need to draw a graph trace, and in general we didn't serialize all that to disk, and even of the stuff we serialized, we don't deserialize it all back in the file_io.Parser classes.

I'm not defending the design — I would now do it differently, perhaps by splitting CSV "metadata" into a block of "customer-facing" fields up at the top, and a block of "technical deets" down at the bottom below the spectra. Or at least ensure that JSON files provide 100% serial/deserial completeness. But that wasn't the priority at the time, and it's not what we now have.

PR #468, opened 2024-08-29
Ensure plugin metadata stomps standard ENLIGHTEN fields in export

mzieg:
A plugin was correctly populating "Declared Match" via the ProcessedReading.plugin_metadata dict, but Measurement.get_metadata was not taking that field from plugin_metadata because the dict key was already found cached in "standard metadata". The checks in get_metadata have been flipped to ensure plugin values "trump" default ENLIGHTEN settings.

### DIFF
--- a/enlighten/measurement/Measurement.py
+++ b/enlighten/measurement/Measurement.py
@@ -815,10 +815,6 @@ class Measurement:
         orig = field
         field = field.lower()
 
-        # if this field has already been computed, use cached
-        if orig in self.metadata:
-            return self.metadata[orig]
-
         # allow plugins to stomp standard metadata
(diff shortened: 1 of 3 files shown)


mzieg:
image

enlighten/measurement/Measurement.py
```diff
@@ -827,6 +823,10 @@ def get_metadata(self, field):
                     log.debug(f"get_metadata: stomping {k} from plugin metadata {v}")
```
jkleinwp, line 823: This output message is a little confusing.  It sounds like the plugin metadata is being stomped

jkleinwp (approved):

PR #520, opened 2024-12-30
Add color matching plugin

jkleinwp:
This PR adds a proof-of-concept color matching plugin to ENLIGHTEN.

The color matching plugin converts a visible spectrum into a color measurement in the color space selected by the user.

To produce reasonable output, a hardcoded calibration factor for a single spectrometer is included.  The calibration can be enabled/disabled via user input. This value will eventually be migrated into and read from EEPROM.

* Convert to RGB, sRGB, color temperature
* Selectable color models (CIE XYZ 1931 2/10 deg and 2005 2/10 deg)
* Plot corrected spectrum
* Enable/disable correction


plugins/RnD/ColorMatching.py
```diff
@@ -0,0 +1,1167 @@
...
+COLOR_MODEL_ALIASES = {
+                        'CIE 1931 2 deg' : 'CIE 1931 2 Degree Standard Observer',
+                        'CIE 1931 10 deg' : 'CIE 1964 10 Degree Standard Observer',
+                        'CIE 2015 2 deg' : 'CIE 2015 2 Degree Standard Observer',
+                        'CIE 2015 10 deg' : 'CIE 2015 10 Degree Standard Observer'
+}
+
+class ColorMatching(EnlightenPluginBase):
+
+        def disconnect(self):
```
mzieg, line 1059: I think you can delete this func, and just use the base class's.

plugins/RnD/ColorMatching.py
```diff
@@ -0,0 +1,1167 @@
...
+                interpolated to 1 nm wavelength spacing (necessary for the color-science package), then converted
+                into CIE XYZ tristimulus values.
+
+                CIE XYZ is the necessary basis for conversion to other colorspaces.
+
+                The function demonstrates conversion to sRGB, RGB and color temperature.
+
+            """
+            wavelength = np.array(self.settings.wavelengths, dtype=np.float64)
+            spectrum = np.array(request.processed_reading.processed, dtype=np.float64)
```
mzieg, line 1128: Consider these for better handling of ProcessedReadings that have already been cropped for horizontal ROI, and/or interpolated by ENLIGHTEN:
```
wavelength = np.array(request.processed_reading.get_wavelengths(), dtype=np.float64)
spectrum   = np.array(request.processed_reading.get_processed(), dtype=np.float64)
```

Note that this would require the calibration array to be similarly altered, as it's in pixel space.

For now, I'd probably just verify that len(spectrum) == len(calibration) before performing the division.

mzieg (approved):

plugins/RnD/ColorMatching.py
jkleinwp: Good call.  Updated to interpolate the calibration factor to the input wavelength range.  Therefore, cropped or otherwise altered spectra will be correctly handled.

plugins/RnD/ColorMatching.py
jkleinwp: Done

PR #530, opened 2025-02-14
Update RPi build instructions

jkleinwp:
Updated build instructions for RPi 4 and 64 bit Bookworm, which does have PySide6 PyPI package.


mzieg (approved):
❤️

Issue #587, opened 2026-05-08
BatchCollection often hangs mid-collection on very short integration times (e.g. 1ms)


mzieg:
Reported by CG and AS. Analysis suggests we're dropping frames in WasatchDeviceWrapper, and so "completing the count" in WasatchDevice before ENLIGHTEN has actually received the requested frame count.

mzieg:
resolved in 4.2.2

PR #592, opened 2026-05-12
4.2.2 dev

mzieg:
- 2026-05-27 4.2.2
    - Infrastructure
        - added EnlightenFeature
            - This was a major architectural change. We previously had a great 
              number of classes loosely called "business objects" (Java term) 
              which had "similar" external interfaces (update_visibility,
              register_observer etc), but they didn't actually utilize 
              inheritance to enforce that commonality. This new ABC provides 
              greater consistency across the growing fleet of ENLIGHTEN feature 
              classes.
            - In the new nomenclature, classes ending with "Feature" should 
              extend EnlightenFeature, and those which don't, should not.
        - ARCHITECTURE docs
        - deprecated timing.Ramp (old utility class)
    - DalaiRamanFeature
        - This was a major goal of this release, to promote DALAI-RAMAN from 
          private plugin to public feature. Note that at present, the .tflite
          models themselves are not stored in this repository. Released
          installers should contain models for "fingerprint" (XM/XS) and 
          "functional" (X) Raman regions.
        - uses alt-graph to RIGHT of Scope
        - show both raw and DALAI on shared clipboard thumbnail
        - save as separate CSV, or combined in JSON
        - displays progress bar during load
    - LibraryMatching
        - Another key goal of this release was to promote LibraryMatching up 
          from plugin status and integrate it as a core feature. 
        - moved standard library to example_data
        - added "[x] use standard library"
        - indicate matches to user-generated library via (*)
        - normalized Pearson score range 
        - deprecated KIA
    - Laser Control
        - IDS Autonomous Collection (key requested feature)
        - BLE laser PWM (key requested feature)
        - ctrl-% shortcut for laser percent
        - simplified password entry
        - add production override (either using production password, or obviating
          due to existing production privs)
    - CalibrationStatusFeature 
        - Another requested feature in this release was this simple on-screen 
          table making it easy to see which of the growing set of corrections and 
          calibrations are in effect.
    - Etalon Correction
        - This new pixel correction improves unit-to-unit library matching
        - should work on XS (EEPROM) or X, XM, IDS, Andor (JSON)
    - InGaAs Correction (experimental)
        - testable on NIR1, 1064 or Andor via JSON
    - Hardware view
        - simplified logic to select and scale product image
        - simplified logic to write EEPROM
        - simplified logic for editing user_text
        - fixed BLE Firmware Rev 
        - fixed PowerWatchdogSec
        - fixed tab order for wavecal coeffs
        - added new Format 19 fields
        - deprecated Baud Rate and Linearity Coeffs
    - refactored FactoryStripChartFeature -> StripChartsFeature
        - deprecated HardwareFileOutputFeature
        - simplified DetectorTemperatureFeature, LaserTemperatureFeature,
                     AmbientTemperatureFeature and BatteryFeature
    - BLEManager
        - add "Last Used" column (laser safety feature)
        - disable under Parallels (avoid sad developers)
    - BatchCollection
        - [#587] fixed "dropped frames" causing BatchCollection to exit early
    - RamanIntensityCorrection
        - on by default when allowed
        - still considering UI changes here, such as eliminating "tri-state"
          on ILP / ILC units and automatically only applying on laser-excited 
          spectra
    - Andor / XL Shutter Control
        - for posterity, before renaming things, empirically validated old 
          behavior was "shutter is OPEN if '[x] Shutter' checkbox is CHECKED" 
          (this was not clear in any way)
        - renaming everything to "Shutter Open" and "Shutter Closed"
        - simplified several other things in AndorDevice, so worth re-testing
    - AccessoryControlXSFeature (UI preview)
        - functionality pending FW
    - Plugins
        - The old ability to move the "other" graph to any position is gone; it 
          is now "locked" to the RIGHT of the Scope Graph, to simplify initial 
          integration with DALAI.
        - Many plugins probably need to be re-tested.
        - removed Raman.LibraryMatching
    - STARTING: 
        - Safe Mode
    - TODO
        - AccessoryControlXSFeature
            - need USB opcodes
        - EtalonCorrection
            - read and cache from EEPROM in BLEDevice
        - BatteryFeature (in work)
            - add state.battery_temperature_deg_c
            - add state.battery_charger_temperature_deg_c
        - StripChartsFeature
            - add legends
        - DidYouKnowFeature
            - update LibraryMatching
            - add DalaiRamanFeature


enlighten/measurement/Measurement.py
```diff
@@ -214,8 +214,9 @@ class Measurement:
                            'Model',
                            'Detector',
                            'Label',
-                           'Declared Match',
-                           'Declared Score',
+                           'Library Match',
+                           'Library Score',
+                           'Library Engine',
```
mzieg, line 219: need to double-check these

enlighten/Plugins/PluginControllerFeature.py
```diff
@@ -1311,24 +1189,21 @@ def clear_previous_layout(self):
...
-        try:
-            if self.plugin_plot is not None:
-                self.plugin_plot.deleteLater()
-                self.plugin_plot = None
-        except:
+        # try:
+        #     if self.plugin_plot is not None:
+        #         self.plugin_plot.deleteLater()
+        #         self.plugin_plot = None
+        # except:
```
mzieg: fix

enlighten/timing/BatchCollectionFeature.py
```diff
@@ -614,6 +616,16 @@ def process(self, pr, spec):
...
 
+        # It looks like we have successfully received a valid Reading which 
+        # should count toward readings_target. Assume this is literally the same
+        # Python object used by WasatchDevice and bump the counter. We are doing
+        # this in ENLIGHTEN rather than Wasatch.PY to account for any "dropped
+        # frames" tossed by WasatchDeviceWrapper. [#587]
+        #
+        # TODO: consider merging TakeOneRequest.readings_current with 
+        # BatchCollectionFeature.current_measurement_count? Meh.
+        tor.readings_current += 1
```
mzieg, line 627: Chris, this was half the fix to your "BatchCollection never completing" issue. (The other half was in Wasatch.PY)

mzieg:
Note this PR is tied to https://github.com/WasatchPhotonics/Wasatch.PY/pull/149

enlighten/measurement/Measurement.py
```diff
@@ -254,8 +257,6 @@ def clear(self):
         self.appending                = False
         self.baseline_correction_algo = None
         self.basename                 = None
-        self.declared_match           = None
-        self.declared_score           = 0
```
graingerchris, line 258: @mzieg   Does Library Match, Library Score, and Library Engine need to be added into this clear as well?  I saw declared_match and declared_score were removed from both the header fields and the clear but those 3 were only added into the header fields.

mzieg: Good spot, yes probably. But the reality is that nothing actually calls most of these clear() methods, and most of them can be removed entirely.

Mostly they were added several years back when we had a weird memory leak that kept crashing Linux systems (it ended up being down in wasatch.applog, and is fixed). But in the meantime, I was being extra-careful on object disposal while trying to figure out what was bloating.

Some clear() methods are called from constructors as a convenient way to initialize all the attributes.

mzieg: On second look, those attributes are actually now held in wasatch.ProcessedReading, and they are in its clear().

mzieg:
This is close enough to release that I'm pushing it to main for final pre-release testing and build.

PR #604, opened 2026-07-20
Cgrainger peek plugin

graingerchris:
A quick plugin to allow peeking at a registry value via i2c.

### DIFF
--- a/plugins/RnD/Peek_Poke_Tester.py
+++ b/plugins/RnD/Peek_Poke_Tester.py
@@ -27,37 +27,22 @@ class Peek_Poke_Tester(EnlightenPluginBase):
         
         self.field(name = "Address", datatype = str, direction = "input", tooltip="values should be entered in hex (leading 0x optional)") 
         self.field(name = "Length", datatype = int, direction = "input", minimum=1, maximum=64, initial=1) 
-        self.field(name = "Value", datatype = str, direction = "output")
-        self.field(name = "Peek", datatype = "button", callback = self.run_peek)
-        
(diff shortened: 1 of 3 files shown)


plugins/RnD/Peek_Poke_Tester.py
```diff
@@ -106,3 +114,62 @@ def get_cmd(self, dev, cmd: int, value: int = 0, index: int = 0, length: int = 6
...
+
+        if not spec.settings.is_xs():
+            self.ctl.marquee.error("PeekPokeTester requires XS")
+            return
+
+        wdw = spec.device # a wasatch.WasatchDeviceWrapper
+        worker = wdw.wrapper_worker # a wasatch.WrapperWorker
+        wasatch_device = worker.connected_device # a wasatch.WasatchDevice (since we already confirmed it's an XS)
+        fid = wasatch_device.hardware # a wasatch.FeatureInterfaceDevice
+        usb_core_device = fid.device # a usb.core.Device from pyusb
```
mzieg: note that everything from this line up could be made a function get_usb_device() and shared with run_peek().

plugins/RnD/Peek_Poke_Tester.py
```diff
@@ -30,6 +30,14 @@ def get_configuration(self):
         self.field(name = "Value", datatype = str, direction = "output")
         self.field(name = "Peek", datatype = "button", callback = self.run_peek)
         
+        self.field(name = "Poke Address", datatype = str, direction = "input", tooltip = "values should be entered in hex (leading 0x optional)")
```
mzieg: Technically I think you could have shared the existing Address field with peek, but okay.

mzieg (approved):

PR #605, opened 2026-07-22
Update to Laser Power Checks

graingerchris:
XS units should have a laser power of 100mW set in the EEPROM.


mzieg (approved):
Nice! Way to extend the metaphor :-)

PR #608, opened 2026-08-04
update to Spectrometer.py to fix a horizontal ROI error

graingerchris:


mzieg (approved):
Looks reasonable, good catch :-)

